Issue #42, opened 2014-05-16
Decompound words


karussell:
One more normalization has to be done to improve search. E.g. Erlangerstraße will be split into "Erlanger straße". This has to be done while indexing and searching.

There is a plugin but it is GPL due to one used library, it could be less restrict but phyton code has to be ported to Java: https://github.com/jprante/elasticsearch-analysis-decompound/issues/5

For POIs there is also often the Bahnhof vs. Hauptbahnhof problem. But probably we should get a main railway station that is not named like this also important in a different country. But probably this should be handled via a different fix: #318

hbruch:
I explored elasticsearch's hyphenation_decompounder. Currently, there exist some issues which need further work:
* ES currently requires an explicit dictionary (word_list) that must contain all subwords to be returned as tokens. Underlying lucene token filter does not, so a custom plugin that instantiates the token filter without word list would work (see [this discussion](https://discuss.elastic.co/t/why-does-hyphenation-decompounder-require-word-list/114567)).
* the hyphenation token filter returns all subwords with offsets identical to the compound word, which results in treating all subwords as synonyms in the query phase. In consequence e.g. searching 'Erlangerstraße' would treat 'erlanger' and 'strasse' as synonyms, which is not intended (see [same discussion](https://discuss.elastic.co/t/why-does-hyphenation-decompounder-require-word-list/114567)). 
* lucene's hyphenation pattern handling has a bit shifting issue (see [LUCENE-8124](https://issues.apache.org/jira/browse/LUCENE-8124)), so patterns are restricted to hyphenation markers 1 to 6. As we'd provide a custom hyphenation pattern file, so this just a minor, avoidable issue.

I'd prepare a WIP PR. The custom plugin should become an artifact on it's own, so what would you recommend: creating a github project on it's own? creating it as maven submodule?

karussell:
Thanks, sounds good to me! I'll then close #46?

Is a separate plugin necessary because of the required internal changes? Or could we utilize it directly in photon somehow?

> creating it as maven submodule?

I would prefer this, yes.

hbruch:
> Thanks, sounds good to me! I'll then close #46?

Yes. I'll reuse the patterns in de.txt and see if applying the decompounder only for *.de.* helps.
 
> Is a separate plugin necessary because of the required internal changes? Or could we utilize it directly in photon somehow?

[...]

PR #46, opened 2014-05-16
initial decompounder settings in separate analyser_de

karussell:
This fixes (partially) #42, is easy to understand and tweak.


karussell:
For Germany this issue is very important to fix! 

If there is no slowdown for the import of other languages, couldn't we enable this just for all languages? Hmmh, I'll investigate also my last comment if we need this only for the query.

hbruch:
@karussell As this PR is rather old, could you answer some questions? 

I assume that using language specific analyzers for name.lang and collector.lang would be ok, as @christophlingg mentioned, wouldn't it?

The [dictionary decompounder](https://www.elastic.co/guide/en/elasticsearch/reference/5.6/analysis-compound-word-tokenfilter.html) came to my mind as well, but is supplying the common endings sufficient? Does it preserve the original term and/or the part preceding the ending? Looking into Lucene's [DictionaryCompoundWordTokenFilter](https://github.com/apache/lucene-solr/blob/master/lucene/analysis/common/src/java/org/apache/lucene/analysis/compound/DictionaryCompoundWordTokenFilter.java) I wonder, but perhaps I'm missing something(?)

Another option (though possibly not scaling to the number of subtokens you had in mind with this PR) could be to use a [Pattern Capture Token Filter](https://www.elastic.co/guide/en/elasticsearch/reference/5.6/analysis-pattern-capture-tokenfilter.html).

Both solutions would require language specific patterns.

A rather language agnostic option might be to switch indexing/querying to ngrams (not just edge_ngrams). But this would result in some major indexing/querying changes and I'm not sure if it would improve overall matching, performance and which effects it had to index size.

karussell:
Sorry do not know what the latest state is also we should add tests when adding it

karussell:
Will be replaced soon.

PR #62, opened 2014-06-05
This supports USA for Photon and Tigerline

mwethington:
This added Tigerline and USA States to Photon

### DIFF
--- a/src/main/java/de/komoot/photon/importer/model/PhotonDoc.java
+++ b/src/main/java/de/komoot/photon/importer/model/PhotonDoc.java
@@ -22,18 +22,19 @@ public class PhotonDoc {
 	final private String tagKey;
 	final private String tagValue;
 	final private Map<String, String> name;
-	final private String houseNumber;
+	private String houseNumber;
 	private String postcode;
 	final private Map<String, String> extratags;
 	final private Envelope bbox;
 	final private long parentPlaceId; // 0 if unset
 	final private double importance;
 	final private CountryCode countryCode;
-	final private Point centroid;
+	private Point centroid;
 	final private long linkedPlaceId; // 0 if unset
 
 	private Map<String, String> street;
 	private Map<String, String> city;
+	private String state;
 	private Set<Map<String, String>> context = new HashSet<Map<String, String>>();
 	private Map<String, String> country;
 
@@ -41,8 +42,7 @@ public class PhotonDoc {
          * Used for testing - really all variables required (final)?         
          */
         public static PhotonDoc create(long placeId, String osmType, long osmId, Map<String, String> nameMap) {
-            return new PhotonDoc(placeId, osmType, osmId, "", "", nameMap, 
-                    "", null, null, 0, 0, null, null, 0);            
+            return new PhotonDoc(placeId, osmType, osmId, "", "", nameMap, null, null, (long)0, 0.0, null, (long)0);
         }
 
 	public boolean isUsefulForIndex() {
--- a/src/main/java/de/komoot/photon/importer/nominatim/NominatimConnector.java
+++ b/src/main/java/de/komoot/photon/importer/nominatim/NominatimConnector.java
@@ -7,6 +7,7 @@ import com.vividsolutions.jts.geom.Point;
 import de.komoot.photon.importer.Importer;
 import de.komoot.photon.importer.model.PhotonDoc;
 import de.komoot.photon.importer.nominatim.model.AddressRow;
+import de.komoot.photon.importer.nominatim.model.TigerRow;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.dbcp.BasicDataSource;
 import org.postgis.jts.JtsWrapper;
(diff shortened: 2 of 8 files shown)


src/main/java/de/komoot/photon/importer/nominatim/NominatimConnector.java
```diff
@@ -96,6 +99,21 @@ public PhotonDoc getByPlaceId(long placeId) {
 		return template.queryForObject("SELECT " + selectColsPlaceX + " FROM placex WHERE place_id = ?", new Object[]{placeId}, placeRowMapper);
 	}
 
+	public List<TigerRow> getTigerAddresses(long placeId) {
+//		placeId = 31210740;
+		return template.query("SELECT place_id, housenumber, postcode, centroid FROM location_property_tiger WHERE parent_place_id = ?", new Object[]{placeId}, new RowMapper<TigerRow>() {
```
christophlingg, line 104: in case a photon user only want's to import osm data (without tiger) will there be an error, maybe the `location_property_tiger` table won't be available.

src/main/java/de/komoot/photon/importer/nominatim/NominatimConnector.java
```diff
@@ -150,11 +169,46 @@ public void processRow(ResultSet rs) throws SQLException {
...
 					}
+
 				}
 
-				if(!doc.isUsefulForIndex()) return; // do not import document
 
-				importer.add(doc);
+				int count = 0;
+//				log.info("place:"+doc.getPlaceId());
+				final List<TigerRow> tigerAddresses = getTigerAddresses(doc.getPlaceId());
```
christophlingg, line 178: thinking about people who don't want to use tiger: for every entry (more than 100 millions) there will be a single db query. what about performance? it already takes a lot of time to import the entire osm data ;-(
I guess it is better to introduce a command line argument to enable tiger imports...

btw. do you know how nominatim handles continuous updates of tiger, do they exist?

src/main/java/de/komoot/photon/importer/model/PhotonDoc.java
```diff
@@ -22,18 +22,19 @@
...
 	final private long parentPlaceId; // 0 if unset
 	final private double importance;
 	final private CountryCode countryCode;
-	final private Point centroid;
+	private Point centroid;
 	final private long linkedPlaceId; // 0 if unset
 
 	private Map<String, String> street;
 	private Map<String, String> city;
+	private String state;
```
christophlingg, line 37: i am not sure about the necessity of state for photon... is it essential for the us?

btw. it should be a map for multilingual support

mwethington, line 37: Yeah map is good for that

You could use province / state for the whole world... States are required for USA

Bill Bell
Sent from mobile

On Jun 16, 2014, at 6:39 AM, "Christoph Lingg" <notifications@github.com<mailto:notifications@github.com>> wrote:

In src/main/java/de/komoot/photon/importer/model/PhotonDoc.java:

> ```
>   final private long linkedPlaceId; // 0 if unset
> 
>   private Map<String, String> street;
>   private Map<String, String> city;
> ```
> -     private String state;

i am not sure about the necessity of state for photon... is it essential for the us?

btw. it should be a map for multilingual support

src/main/java/de/komoot/photon/importer/nominatim/NominatimConnector.java
mwethington, line 178: Tiger option is good... There is no easy way to continuous update. Tiger is updated once a year by the govt...

But in USA the Osm street data is sparse and Tiger is needed....

Bill Bell
Sent from mobile

On Jun 16, 2014, at 6:36 AM, "Christoph Lingg" <notifications@github.com<mailto:notifications@github.com>> wrote:

In src/main/java/de/komoot/photon/importer/nominatim/NominatimConnector.java:

> -                             importer.add(doc);
> -                             int count = 0;
>   +//                           log.info("place:"+doc.getPlaceId());
> -                             final List<TigerRow> tigerAddresses = getTigerAddresses(doc.getPlaceId());

thinking about people who don't want to use tiger: for every entry (more than 100 millions) there will be a single db query. what about performance? it already takes a lot of time to import the entire osm data ;-(
I guess it is better to introduce a command line argument to enable tiger imports...

btw. do you know how nominatim handles continuous updates of tiger, do they exist?

src/main/java/de/komoot/photon/importer/nominatim/NominatimConnector.java
mwethington, line 104: Yeah option is good

Bill Bell
Sent from mobile

On Jun 16, 2014, at 6:32 AM, "Christoph Lingg" <notifications@github.com<mailto:notifications@github.com>> wrote:

In src/main/java/de/komoot/photon/importer/nominatim/NominatimConnector.java:

> @@ -96,6 +99,21 @@ public PhotonDoc getByPlaceId(long placeId) {
>               return template.queryForObject("SELECT " + selectColsPlaceX + " FROM placex WHERE place_id = ?", new Object[]{placeId}, placeRowMapper);
>       }
> -     public List<TigerRow> getTigerAddresses(long placeId) {
>   +//           placeId = 31210740;
> -             return template.query("SELECT place_id, housenumber, postcode, centroid FROM location_property_tiger WHERE parent_place_id = ?", new Object[]{placeId}, new RowMapper<TigerRow>() {

in case a photon user only want's to import osm data (without tiger) will there be an error, maybe the location_property_tiger table won't be available.

karussell:
Any intentions to address the comments @mwethington? What happens if there is a property in the Tiger data and in nominatim - will they be in elasticsearch as duplicates? How could we avoid this? Or was the plan to use only Tiger in the US?

lonvia:
I don't think it is worth keeping this PR open. The format of the table that contains the Tiger data has  changed significantly since this PR was made. Tiger house numbers are now interpolated on the fly. So this import code will need a thorough rewrite.

karussell:
Tiger data is already in nominatim?

lonvia:
Yes, Tiger house numbers can optionally be used (they are currently not in the database that is producing the weekly photon dumps).

Issue #65, opened 2014-06-10
test fails


karussell:
QueryParsingException[[photon] script_score the script could not be loaded

Where is this script?

PR #66, opened 2014-06-10
fixed snapshotting

karussell:
There were several issues with the current snapshotting like GetSnapshot vs. RestoreSnapshot - some of them only jumped to my eyes when looking at the [ES tests](https://github.com/elasticsearch/elasticsearch/blob/master/src/test/java/org/elasticsearch/snapshots/SharedClusterSnapshotRestoreTests.java#L126). Also moved some log.error statements into exceptions as it does not make sense to continue in those cases.

There is an additional test for it now but the whole test suite does not execute properly - depends on #65

Upgraded to elasticsearch as ES docs say: *Before version 1.2.0 the snapshot operation fails if cluster has any relocating or initializing primaries of indices participating in the snapshot. Starting with version 1.2.0, Elasticsearch waits for are relocating or initializing shards to start before snapshotting them.*

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -10,7 +10,7 @@
 		<dependency>
 			<groupId>org.elasticsearch</groupId>
 			<artifactId>elasticsearch</artifactId>
-			<version>1.1.0</version>
+			<version>1.2.0</version>
 		</dependency>
 		<dependency>
 			<groupId>postgresql</groupId>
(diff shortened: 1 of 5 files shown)


karussell:
I think we can close here and use elasticsearch/curator instead, right?

christophlingg:
right

PR #83, opened 2014-08-03
Merge with positive scoring

yohanboniface:
WIP

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,13 +22,11 @@ photon requires java, at least version 6.
 
 get photon
 ```bash
-wget http://photon.komoot.de/data/photon-0.1.tar.gz
-tar -zxvf data.tar.gz
+wget http://photon.komoot.de/data/photon-0.1.jar
 ```
 
 download search index (x,x gb, worldwide, languages: English, German, French and Italian)
  ```bash
-cd photon-0.1
 wget http://photon.komoot.de/data/photon_data_world_5-07-2014.tar.bz2
 tar xvjf photon_data_world_5-07-2014.tar.bz2
 ln -s photon_data photon_data_world_5-07-2014.tar.bz2


es/query.json
```diff
@@ -1,34 +1,78 @@
...
+        "filters": [
+          {
+            "missing": {
+              "field": "housenumber"
+            }
+          },
+          {
+            "query": {
+              "match": {
+                "housenumber": {
```
christophlingg, line 80: I removed this filter in the komoot branch, it was notable time consuming, in particular if you search for very popular terms. Also, alphanumeric housenumber like 25c did not work with this filter. as a tokenizer was splitting it into [`25`, `c`] there was not match any more. do you know how to fix it? maybe we don't need it at all, i haven't found a situation where it caused problems. however, in terms of logic, this filter is valid.

maybe

es/query_location_bias.json
```diff
@@ -1,40 +1,88 @@
...
           {
             "script_score": {
-              "params": {"lat": ${lat}, "maxDist": 100, "lon": ${lon}},
-              "script": "dist = doc['coordinate'].distanceInKm(lat, lon); 1 / (0.5 - 0.5 * exp(-5*dist/maxDist))"
+              "script": "1 + doc['importance'].value * 100"
+            }
+          },
+          {
+            "script_score": {
+              "script": "dist = doc['coordinate'].distanceInKm(lat, lon); 1 / (0.5 - 0.5 * exp(-5*dist/maxDist))",
```
christophlingg: i found out that using a ration function instead of an exponential function speed up location based searches by about 20 %. the evolution of the graph can be very very similar, i'll make a commit for it later.

es/query.json
yohanboniface, line 80: Not sure to understand which filter you have removed. All the housenumber handling, or the query from line 60 to 68?

es/query_location_bias.json
yohanboniface: Great! I'm happy to let the physician of us play with this :)

PR #100, opened 2014-08-18
Enable running on Windows

TheRealPanPan:
Try to setupDirectories using "file:///" protocol when "file://" fails on windows system due to missing authoritative component of the URI.

### DIFF
--- a/src/main/java/de/komoot/photon/importer/elasticsearch/Server.java
+++ b/src/main/java/de/komoot/photon/importer/elasticsearch/Server.java
@@ -52,6 +52,13 @@ public class Server {
 			setupDirectories(new URL("file://" + mainDirectory));
 		} catch(MalformedURLException e) {
 			log.error("Can´t create directories");
+		} catch(Exception ex)
+		{
+			try {
+				setupDirectories(new URL("file:///" + mainDirectory)); //Enable running on windows.
+			} catch(MalformedURLException e) {
+				log.error("Can´t create directories");
+			}
 		}
 	}
 


src/main/java/de/komoot/photon/importer/elasticsearch/Server.java
```diff
@@ -52,6 +52,13 @@ public Server(String mainDirectory) {
 			setupDirectories(new URL("file://" + mainDirectory));
 		} catch(MalformedURLException e) {
 			log.error("Can´t create directories");
+		} catch(Exception ex)
+		{
+			try {
+				setupDirectories(new URL("file:///" + mainDirectory)); //Enable running on windows.
```
christophlingg, line 58: ins't there a more general / elegant way to do this in java? have you had a look on alternatives, maybe `file:///` works on unix systems too. it's not  a very big deal, but i guess there is a nicer solution than catching exceptions twice...

TheRealPanPan, line 58: Don't really know if it can be done in a more elegant way.
The Exception that is raised is : 
Exception in thread "main" java.lang.IllegalArgumentException: URI has an authority component
    at java.io.File.<init>(Unknown Source)
    at de.komoot.photon.importer.elasticsearch.Server.setupDirectories(Server.java:231)
    at de.komoot.photon.importer.elasticsearch.Server.<init>(Server.java:52)
    at de.komoot.photon.importer.App.main(App.java:68)

After two - three minutes of searching with my friends google I find a lot of these kind of solution. As i'm not a Java dev (much more of a C/C++ one) i've posted this fix in order to raise awarness of the existing problem and having an easy way to bypass it.

I don't know if file:/// will work on linux. I didn't try it :/

PR #125, opened 2014-12-10
Making languages dynamic with command arguments

Svantulden:
**Heya!** 

Thanks for this amazing library! Photon is best suited for our geocoding autocomplete needs compared to Nominatim (which we still use for reverse-geocoding and details). 

I made some changes to allow for dynamic languages in Photon. We specifically needed to add Dutch as a language, and while I went over the language-specific code in Photon, I noticed that it wouldn't be too hard to make it fully dynamic. 

The way it works now is as follows: 
### To specify languages that should be imported during nominatim-import

Use the `nominatim-import-languages` command argument. i.e.
`java -jar photon-0.1.2.jar -nominatim-import -nominatim-import-languages nl,en -host localhost -port 5432 -database nominatim -user user -password password` 
for importing only the default, en and nl tags.

This could also be handy to keep Photon data as small as possible without languages that you are sure you are not going to use.
### To specify languages for run-time

This could maybe be handy to block languages/enable languages for the request handler. Use the `supported-languages` command argument. i.e.
`java -jar photon-0.1.2.jar -supported-languages nl,en`
to use the nl and en language tags (defaults to en when no lang is specified).

**However...**
When testing this functionality, I noticed that the order of results was different between `lang=en` and `lang=nl`. After some debugging I noticed that it was the same on all the previous enabled languages `(en,fr,it,de)`, but different on the `nl` tag. 

I found out this was because I didn't update the `mappings.json` which is used on import, to specify collectors for languages. This should be dynamically changed too, when specifying `nominatim-import-languages`, which leaves this PR **not finished**. 
### Thoughts?

Thoughts & discussions are very welcome, on the code I wrote and on how I/we could tackle this problem correctly. 
### Things to do:
- [x] Make mappings.json dynamic somehow
- [x] Fully test the changes
- [x] How would we handle the problem when `supported-languages` has specified a language that doesn't exist in the data

**PS.** I also added specific Dutch postcode checking to this PR, which may help other Dutch users of Photon. This filters out double `highway` names who are part of the same street. In The Netherlands, a street can have different postcode letters (i.e. 6532RA and 6532BE could be the same street). The numbers however, stay the same. So to avoid clutter, I filter results that have the same postcode number AND name.

### DIFF
--- a/src/main/java/de/komoot/photon/importer/elasticsearch/Server.java
+++ b/src/main/java/de/komoot/photon/importer/elasticsearch/Server.java
@@ -168,45 +168,44 @@ public class Server {
                 String nameToCollectorString = "{\"type\":\"string\",\"index\":\"no\",\"fields\":{\"ngrams\":{\"type\":\"string\",\"index_analyzer\":\"index_ngram\"},\"raw\":{\"type\":\"string\",\"index_analyzer\":\"index_raw\"}},\"copy_to\":[\"collector.{lang}\"]}";
                 String collectorString = "{\"type\":\"string\",\"index\":\"no\",\"fields\":{\"ngrams\":{\"type\":\"string\",\"index_analyzer\":\"index_ngram\"},\"raw\":{\"type\":\"string\",\"index_analyzer\":\"index_raw\"}},\"copy_to\":[\"collector.{lang}\"]}}},\"street\":{\"type\":\"object\",\"properties\":{\"default\":{\"index\":\"no\",\"type\":\"string\",\"copy_to\":[\"collector.default\"]}";
                 
-                if(mappingsObject != null) {
-                        JSONObject placeObject = mappingsObject.optJSONObject("place");
-                        JSONObject propertiesObject = placeObject == null ? null : placeObject.optJSONObject("properties");
-                        
-                        if(propertiesObject != null) {
-                                for(String lang : langs) {
-                                        // create lang-specific json objects
-                                        JSONObject copyToCollectorObject = new JSONObject(copyToCollectorString.replace("{lang}", lang));
-                                        JSONObject nameToCollectorObject = new JSONObject(nameToCollectorString.replace("{lang}", lang));
-                                        JSONObject collectorObject = new JSONObject(collectorString.replace("{lang}", lang));
-                                    
-                                        // add language specific tags to the collector
-                                        propertiesObject = addToCollector("city", propertiesObject, copyToCollectorObject, lang);
-                                        propertiesObject = addToCollector("context", propertiesObject, copyToCollectorObject, lang);
-                                        propertiesObject = addToCollector("country", propertiesObject, copyToCollectorObject, lang);
-                                        propertiesObject = addToCollector("street", propertiesObject, copyToCollectorObject, lang);
-                                        propertiesObject = addToCollector("name", propertiesObject, nameToCollectorObject, lang);
-                                        
-                                        // add language specific collector to default for name
-                                        JSONObject name = propertiesObject.optJSONObject("name");
-                                        JSONObject nameProperties = name == null ? null : name.optJSONObject("properties");
-                                        if(nameProperties != null) {
-                                                JSONObject defaultObject = nameProperties.optJSONObject("default");
-                                                JSONArray copyToArray = defaultObject == null ? null : defaultObject.optJSONArray("copy_to");
-                                                copyToArray.put("name." + lang);
-                                                
-                                                defaultObject.put("copy_to", copyToArray);
-                                                nameProperties.put("default", defaultObject);
-                                                name.put("properties", nameProperties);
-                                                propertiesObject.put("name", name);
-                                        }
-                                        
-                                        // add language specific collector
-                                        propertiesObject = addToCollector("collector", propertiesObject, collectorObject, lang);                                        
+                JSONObject placeObject = mappingsObject.optJSONObject("place");
+                JSONObject propertiesObject = placeObject == null ? null : placeObject.optJSONObject("properties");
+
+                if(propertiesObject != null) {
+                        for(String lang : langs) {
+                                // create lang-specific json objects
+                                JSONObject copyToCollectorObject = new JSONObject(copyToCollectorString.replace("{lang}", lang));
+                                JSONObject nameToCollectorObject = new JSONObject(nameToCollectorString.replace("{lang}", lang));
+                                JSONObject collectorObject = new JSONObject(collectorString.replace("{lang}", lang));
+
+                                // add language specific tags to the collector
+                                propertiesObject = addToCollector("city", propertiesObject, copyToCollectorObject, lang);
+                                propertiesObject = addToCollector("context", propertiesObject, copyToCollectorObject, lang);
+                                propertiesObject = addToCollector("country", propertiesObject, copyToCollectorObject, lang);
+                                propertiesObject = addToCollector("street", propertiesObject, copyToCollectorObject, lang);
+                                propertiesObject = addToCollector("name", propertiesObject, nameToCollectorObject, lang);
+
+                                // add language specific collector to default for name
+                                JSONObject name = propertiesObject.optJSONObject("name");
+                                JSONObject nameProperties = name == null ? null : name.optJSONObject("properties");
+                                if(nameProperties != null) {
+                                        JSONObject defaultObject = nameProperties.optJSONObject("default");
+                                        JSONArray copyToArray = defaultObject == null ? null : defaultObject.optJSONArray("copy_to");
+                                        copyToArray.put("name." + lang);
+
+                                        defaultObject.put("copy_to", copyToArray);
+                                        nameProperties.put("default", defaultObject);
+                                        name.put("properties", nameProperties);
+                                        propertiesObject.put("name", name);
                                 }
-                                placeObject.put("properties", propertiesObject);
-                                return mappingsObject.put("place", placeObject);
+
+                                // add language specific collector
+                                propertiesObject = addToCollector("collector", propertiesObject, collectorObject, lang);                                        
                         }
+                        placeObject.put("properties", propertiesObject);
+                        return mappingsObject.put("place", placeObject);
                 }
+                
                 log.error("cannot add langs to mapping.json, please double-check the mappings.json or the language values supplied");
                 return null;
         }


es/query_location_bias.json
```diff
@@ -67,8 +67,8 @@
             "script_score": {
               "script": "dist = doc['coordinate'].distanceInKm(lat, lon); 0.5 + ( 1.5 / (1.0 + dist/40.0) )",
               "params": {
-                "lat": ${lat},
-                "lon": ${lon}
+                "lat": "${lat}",
```
yohanboniface: This will cast floats to strings, not sure we really want to do that. What was your rationale here (apart from making the json a vaild one ;) )?

src/main/java/de/komoot/photon/importer/RequestHandler.java
```diff
@@ -19,11 +20,12 @@
  */
 public class RequestHandler extends Route {
 	private final Searcher searcher;
-	private static final Set<String> supportedLanguages = ImmutableSet.of("de", "en", "fr", "it");
-
-	protected RequestHandler(String path, Searcher searcher) {
+	private final Set<String> supportedLanguages;
+
```
yohanboniface: I see spaces here, while the lines around have tabs, but I'm not sure what's the choice here. @christophlingg? :)

src/main/java/de/komoot/photon/importer/elasticsearch/Searcher.java
```diff
@@ -84,12 +85,35 @@ public Searcher(Client client) {
...
 					String postcode = properties.getString(OSMTags.KEY_POSTCODE);
 					String name = properties.getString(OSMTags.KEY_NAME);
-					String key = postcode + ":" + name;
+					String key = postcode + ":" + name;                                                                   
 
 					if(keys.contains(key)) {
 						// a street with this name + postcode is already part of the result list
 						continue;
-					}
+					} else if (lang.equals("nl")) {
```
yohanboniface: I think this should better land in a separate PR.

src/main/java/de/komoot/photon/importer/CommandLineArgs.java
```diff
@@ -16,6 +16,12 @@
 	
 	@Parameter(names = "-nominatim-import", description = "import nominatim database into photon (this will delete previous index)")
 	private boolean nominatimImport = false;
+        
+        @Parameter(names = "-nominatim-import-languages", description = "languages nominatim importer should import, comma seperated (default: 'en,fr,de,it')")
```
christophlingg: can we unify both arguments? could be cleaner

es/query_location_bias.json
Svantulden: Woops! That is an oversight on my part.. Did not realize it would be casted as string. Will revert this change

src/main/java/de/komoot/photon/importer/RequestHandler.java
Svantulden: Probably a formatting setting of the Netbeans config I use. I will change it to tabs.

src/main/java/de/komoot/photon/importer/elasticsearch/Searcher.java
Svantulden: Yeah I wasn't sure about having this in the same PR. Because it is dependent of the previous changes from this PR, I thought I would add it here. Will submit this as a separate PR :+1:

src/main/java/de/komoot/photon/importer/CommandLineArgs.java
Svantulden: That's a good idea. I was struggling with a unified name, what do you suggest? Something like '-using-languages' or just '-supported-languages' maybe?

src/main/java/de/komoot/photon/importer/CommandLineArgs.java
```diff
@@ -16,6 +16,9 @@
 	
 	@Parameter(names = "-nominatim-import", description = "import nominatim database into photon (this will delete previous index)")
 	private boolean nominatimImport = false;
+        
+        @Parameter(names = "-using-languages", description = "languages nominatim importer should import and use at run-time, comma seperated (default: 'en,fr,de,it')")
```
yohanboniface: What about just `-languages`?

src/main/java/de/komoot/photon/importer/RequestHandler.java
yohanboniface: Spaces still seem to be there ;)

src/main/java/de/komoot/photon/importer/Utils.java
```diff
@@ -22,9 +22,9 @@
 public class Utils {
 	static final Joiner commaJoiner = Joiner.on(", ").skipNulls();
 
-	final static String[] languages = new String[]{"de", "en", "fr", "it"};
+	//final static String[] languages = new String[]{"de", "en", "fr", "it"};
```
yohanboniface: If we don't need this line, let's remove it :)

src/main/java/de/komoot/photon/importer/elasticsearch/Server.java
```diff
@@ -141,8 +145,15 @@ public void recreateIndex() {
 		final InputStream index_settings = Thread.currentThread().getContextClassLoader().getResourceAsStream("index_settings.json");
 
 		try {
+                        // get mappings as JSONObject
+                        String mappingsString = IOUtils.toString(mappings);
```
yohanboniface: I generally prefer to have in the `try` block only the line that should actually raise.

src/main/java/de/komoot/photon/importer/elasticsearch/Server.java
```diff
@@ -156,4 +167,91 @@ public DeleteIndexResponse deleteIndex() {
...
+                        
+                        if(propertiesObject != null) {
+                                for(String lang : langs) {
+                                        // create lang-specific json objects
+                                        JSONObject copyToCollectorObject = new JSONObject(copyToCollectorString.replace("{lang}", lang));
+                                        JSONObject nameToCollectorObject = new JSONObject(nameToCollectorString.replace("{lang}", lang));
+                                        JSONObject collectorObject = new JSONObject(collectorString.replace("{lang}", lang));
+                                    
+                                        // add language specific city to the collector
+                                        JSONObject city = propertiesObject.optJSONObject("city");
```
yohanboniface: This may worth a bit of factorizing, yep. Something like `addToCollector(key, propertiesObject)` should do the trick, nope?

src/main/java/de/komoot/photon/importer/CommandLineArgs.java
Svantulden: Simple and to the point, much better :+1:

src/main/java/de/komoot/photon/importer/Utils.java
Svantulden: Thanks for catching that, a bit sloppy of me

src/main/java/de/komoot/photon/importer/elasticsearch/Server.java
Svantulden: Yeah, I reworked it for all except the name default, because that needs some custom adding. Thanks

src/main/java/de/komoot/photon/importer/elasticsearch/Server.java
Svantulden: That results in a bit of unclean code i.m.o. but I see what you are saying. I agree it would be better to be consistent with that.

src/main/java/de/komoot/photon/importer/elasticsearch/Server.java
```diff
@@ -139,10 +143,21 @@ public void recreateIndex() {
...
 		final InputStream mappings = Thread.currentThread().getContextClassLoader().getResourceAsStream("mappings.json");
 		final InputStream index_settings = Thread.currentThread().getContextClassLoader().getResourceAsStream("index_settings.json");
-
+                
+                String mappingsString = "";
 		try {
+                        // get mappings as JSONObject
+                        mappingsString = IOUtils.toString(mappings);
+                } catch(IOException e) {
+			log.error("cannot setup index, elastic search config files not readable", e);
```
christophlingg: I would rethrow the exception here... otherwise photon would start with a broken mapping and it will break anyway with strange message that make it difficult to understand what's broken. I think it's a good approach to be strict here.

src/main/java/de/komoot/photon/importer/elasticsearch/Server.java
```diff
@@ -139,10 +143,21 @@ public void recreateIndex() {
...
+                JSONObject mappingsJSON = new JSONObject(mappingsString);
+                        
+                // add all langs to the mapping
+                mappingsJSON = addLangsToMapping(mappingsJSON);
+                try {        
 			client.admin().indices().prepareCreate("photon").setSettings(IOUtils.toString(index_settings)).execute().actionGet();
-			client.admin().indices().preparePutMapping("photon").setType("place").setSource(IOUtils.toString(mappings)).execute().actionGet();
+			client.admin().indices().preparePutMapping("photon").setType("place").setSource(mappingsJSON.toString()).execute().actionGet();
 		} catch(IOException e) {
 			log.error("cannot setup index, elastic search config files not readable", e);
```
christophlingg: same here

src/main/java/de/komoot/photon/importer/elasticsearch/Server.java
```diff
@@ -156,4 +171,64 @@ public DeleteIndexResponse deleteIndex() {
...
 		}
 	}
+        
+        private JSONObject addLangsToMapping(JSONObject mappingsObject) {
+                // define collector json strings
+                String copyToCollectorString = "{\"type\":\"string\",\"index\":\"no\",\"copy_to\":[\"collector.{lang}\"]}";
+                String nameToCollectorString = "{\"type\":\"string\",\"index\":\"no\",\"fields\":{\"ngrams\":{\"type\":\"string\",\"index_analyzer\":\"index_ngram\"},\"raw\":{\"type\":\"string\",\"index_analyzer\":\"index_raw\"}},\"copy_to\":[\"collector.{lang}\"]}";
+                String collectorString = "{\"type\":\"string\",\"index\":\"no\",\"fields\":{\"ngrams\":{\"type\":\"string\",\"index_analyzer\":\"index_ngram\"},\"raw\":{\"type\":\"string\",\"index_analyzer\":\"index_raw\"}},\"copy_to\":[\"collector.{lang}\"]}}},\"street\":{\"type\":\"object\",\"properties\":{\"default\":{\"index\":\"no\",\"type\":\"string\",\"copy_to\":[\"collector.default\"]}";
+                
+                if(mappingsObject != null) {
```
christophlingg: can it be null here? if you are stricter above, maybe you don't need to care about nulls anymore

src/main/java/de/komoot/photon/importer/elasticsearch/Server.java
```diff
@@ -156,4 +171,64 @@ public DeleteIndexResponse deleteIndex() {
 			return null;
 		}
 	}
+        
+        private JSONObject addLangsToMapping(JSONObject mappingsObject) {
+                // define collector json strings
+                String copyToCollectorString = "{\"type\":\"string\",\"index\":\"no\",\"copy_to\":[\"collector.{lang}\"]}";
```
christophlingg, line 167: you are dealing with strings here, which is ok. I would have investigated if there is a way circumvent building json with string templates but with the `org.json` package

src/main/java/de/komoot/photon/importer/elasticsearch/Updater.java
```diff
@@ -18,10 +18,12 @@
 public class Updater implements de.komoot.photon.importer.Updater {
 	private Client esClient;
 	private BulkRequestBuilder bulkRequest;
+        private String[] languages;
```
christophlingg: you can make this final, so you are sure nobody can change it after class has been instanced

src/main/java/de/komoot/photon/importer/json/JsonDumper.java
```diff
@@ -17,16 +17,18 @@
 @Slf4j
 public class JsonDumper implements Importer {
 	private PrintWriter writer = null;
+        private String[] languages;
```
christophlingg: same with final here

src/main/java/de/komoot/photon/importer/elasticsearch/Server.java
Svantulden: Hmm, wouldn't it be bad practice to rethrow the exception? (see [here](http://howtodoinjava.com/2013/04/04/java-exception-handling-best-practices/)) 

I reworked some of the exception handling regarding the index creating (commit 1a50011), so that the program will not continue and properly throw with the stacktrace. The tests will also fail due to the exception throwing. I would like your opinion on this.

src/main/java/de/komoot/photon/importer/elasticsearch/Updater.java
Svantulden: Missed those ones, thanks!

src/main/java/de/komoot/photon/importer/elasticsearch/Server.java
Svantulden, line 167: Yeah I thought about it to purely use the json package, which would be more consistent, but it would in my opinion be a bad effort/reward situation. It's easy to replace strings with the lang, while it would be cumbersome to do so with navigating through the json objects. Also, when the mappings change, much would have to be rewritten.

PR #128, opened 2014-12-22
Remove street duplicates for NL postcodes

Svantulden:
Hi, 

This PR could help people who use Photon to search for Dutch addresses. It has a specific check for NL postcodes, that results in removing duplicate street names. Doing so solves this problem:

image

Here we get 5 of the same results for the query "veldsingel", which could block other queries called "veldsingel" in another town/city. These are all part of the same street, but don't get filtered automatically by the `removeStreetDuplicates` function. This is because it assumes a postcode only consists of numbers, while in The Netherlands, it consists of 4 numbers and 2 letters. The same street can have the same numbers but different letters. 

This extra check, ensures we only get one street result per street, effectively removing the duplicates based on it having the same 4 numbers and having the same streetname.

### DIFF
--- a/src/main/java/de/komoot/photon/importer/elasticsearch/Searcher.java
+++ b/src/main/java/de/komoot/photon/importer/elasticsearch/Searcher.java
@@ -84,47 +84,19 @@ public class Searcher {
 				if(properties.has(OSMTags.KEY_POSTCODE) && properties.has(OSMTags.KEY_NAME)) {
 					// street has a postcode and name
 					String postcode = properties.getString(OSMTags.KEY_POSTCODE);
+                                        String onlyDigitsPostcode = Utils.stripNonDigits(postcode);
 					String name = properties.getString(OSMTags.KEY_NAME);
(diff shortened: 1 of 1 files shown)


src/main/java/de/komoot/photon/importer/elasticsearch/Searcher.java
```diff
@@ -90,7 +90,42 @@ public Searcher(Client client) {
 					if(keys.contains(key)) {
 						// a street with this name + postcode is already part of the result list
 						continue;
-					} 
+					} else if (lang.equals("nl")) {
```
christophlingg: isn't there a way to save a lot of lines of code by using a language-specific key generation? let's say:

```
if lang == "nl":
    key = onlyDigitsPostCode + ':' + 'name'
else:
    key = postcode + ":" + name;    
```

Svantulden: Woops.. Massive oversight on my part, thanks for catching this!

PR #129, opened 2014-12-22
Using optional tag-->values whitelist json file during import

Svantulden:
Hi, 

For our use case, I needed to filter out some of the POI's and other unneeded tags from the photon data. To do this, instead of specifying which tags to filter out (which there are loads of, adding new ones every day), I made a json file that holds a whitelist for tags and it's values. 

I made this PR because I think other people could benefit from this too. By using the json tag whitelist (I've added an example json file outside of the project, to help with the format), it has the following pro's: 
- Significantly reduces file size for the photon_data. By using the example json file (which we use in our project), we reduced the size from 15GB to 3GB. 
- Get more reliable data in photon, by knowing which tags and values can be returned
- Query lookup should be faster (I haven't tested how much faster it can be, but I assume it benefits from having a smaller data table to look through, though I am not sure how it affects ES timings)

Feedback is welcome! Would this be too much of a specific use case for Photon to merge?

### DIFF
--- a/tags_whitelist_example.json
+++ b/tags_whitelist_example.json
@@ -5,6 +5,7 @@
       "hamlet",
       "island",
       "town",
+      "village",
       "isolated_dwelling",
       "suburb",
       "islet",


Svantulden:
I understand, thanks for looking at it though!

Issue #68, opened 2014-06-11
Allow searching on osm_key and osm_value


karussell:
There should be another filter parameter in the API to search in a specific area (bounding box) for a specific tag combination or even a set of combinations ('I want food').

For that osm_key and osm_value have to be indexed in the mapping. Or maybe we make this configurable if this increases size too much.

sdole:
I am working on this issue. I added a few new [PhotonDoc](https://github.com/sdole/photon/blob/master/src/main/java/de/komoot/photon/PhotonDoc.java)s into the setup of the [QueryStateTest](https://github.com/sdole/photon/blob/master/src/test/java/de/komoot/photon/elasticsearch/QueryStateTest.java). These new PhotonDocs have multiple arbitrary tags in the "extratags" constructor parameter. I expected these extratags to be in the result on [line 52](https://github.com/sdole/photon/blob/master/src/test/java/de/komoot/photon/elasticsearch/QueryStateTest.java#L52) - but none showed up. What am I doing wrong? Or - is the test index not setup to have the extra tags?

Please let me know if my question does not make sense. 

Thanks

sdole:
I committed [an example on my fork](https://github.com/sdole/photon/blob/master/src/test/java/de/komoot/photon/elasticsearch/TagFilterTest.java). See [comments on that commit](16b0a33b4230eee7a51a7b895aadb09b15872cbe).

christophlingg:
In extratags we save some additional data that is coming from nominatim's db column, here is an example place: http://nominatim.openstreetmap.org/details.php?place_id=167287581

We added it in the beginning but we have never used it so far.

We should be very careful about what information we store in the elasticsearch index. We have more than 100.000.000 documents, every little detail we add will have a big impact on the index size. The information in extratags are not important enough to be stored.

If you want to save this information for your needs, you need to adopt this util function: https://github.com/komoot/photon/blob/master/src/main/java/de/komoot/photon/Utils.java#L24 It defines how a photon doc is converted into a elasticsearch doc.

Btw: you do not need extratags to allow filtering on osm_key and osm_value

karussell:
Maybe there is an option where one can enable this functionality?

sdole:
My mistake. The correct questions I should have asked are:

1. Where are osm tags stored in the ES data model? and
2. Do you have suggestions on how I should formulate my es query for
filtering on tags.

Thank you.

christophlingg:
[...]

PR #141, opened 2015-02-06
fixes issue 68

sdole:
This is my attempt to fix komoot/photon#68.

With this commit, 
1. we can add two new url parameters "tagKey" and "tagValue".
2. both these are optional, however, if tagValue is provided, then tagKey must be provided.
3. if neither is provided, then no filtering is done on search results
4. if tagKey is provided (such as "tourism"), then only those places that have a tag with that key ("tourism") will be returned (everything else will be filtered).
5. if tagValue is additionally provided (such as "museum"), then, only those places that have a tag with that key ("tourism") and that tagValue ("museum") will be returned.

Let me know if you have questions or concerns.


karussell:
This looks nice! Maybe indexing of these attributes should be made configurable and disabled by default to avoid big indices?

es/mappings.json
```diff
@@ -16,7 +16,7 @@
       },
       "osm_key": {
         "type": "string",
-        "index": "no"
+        "index":"not_analyzed"
```
christophlingg, line 19: we should add `not_analyzed` to osm_value too. at the moment it is on default: `analyzed` which is not what we want http://www.elasticsearch.org/guide/en/elasticsearch/guide/current/mapping-intro.html

sdole, line 19: I can do that. I noticed it, but didn't want to change what's already
there. I will do that today.

christophlingg:
great stuff, i am going through the changes and adding some thoughts!

christophlingg:
before we had two json files, now there are six. When we want to change the query we will have to it for every file which is quite error prone. Probably it would be clearer if we would use a very simple template engine that provides `if` /  `else` statements that can be used to make one single query file.

There is a stackoverflow discussion about lightweight template engines for java: http://stackoverflow.com/questions/3793880/lightweight-template-engine-in-java

a [mustache based engine](http://stackoverflow.com/a/14623081/1245622) seems to be a good option: https://github.com/spullara/mustache.java

src/main/java/de/komoot/photon/OsmTag.java
```diff
@@ -0,0 +1,22 @@
+package de.komoot.photon;
+
+/**
+ * Created by Sachin Dole on 2/6/2015.
+ */
+public class OsmTag {
```
christophlingg, line 6: if you are lazy you can annotate the class with `@Data` from lombok  that writes setter and getter as well as the constructor automatically.

src/main/java/de/komoot/photon/elasticsearch/Searcher.java
```diff
@@ -57,12 +71,32 @@ public Searcher(Client client) {
 		if(lon != null) params.put("lon", lon);
 		if(lat != null) params.put("lat", lat);
 
-		StrSubstitutor sub = new StrSubstitutor(params.build(), "${", "}");
-		if(lon != null && lat != null) {
-			query = sub.replace(queryLocationBiasTemplate);
-		} else {
-			query = sub.replace(queryTemplate);
-		}
+        this.setupSearch(lon,lat,tagKey,tagValue);
```
christophlingg, line 74: this battery of if/else statements could be replaced by using a template engine and only one json template

src/main/java/de/komoot/photon/RequestHandler.java
```diff
@@ -59,10 +59,13 @@ public String handle(Request request, Response response) {
 			limit = 15;
 		}
 
-		List<JSONObject> results = searcher.search(query, lang, lon, lat, limit, true);
+        String tagKey = request.queryParams("tagKey");
+        String tagValue = request.queryParams("tagValue");
```
christophlingg: ok, this is a difficult issue as there is no real convention for url parameters. camelcase should be used in java application. However when it comes to urls or query parameters i would rather use underscore variants: `tag_key` & `tag_value` identical as it was defined in our [mapping.json](https://github.com/komoot/photon/blob/master/es/mappings.json#L25)

But as I said, there is no real convention, see
- http://stackoverflow.com/questions/10302179/hyphen-underscore-or-camelcase-as-word-delimiter-in-uris
- http://stackoverflow.com/questions/778203/are-there-any-naming-convention-guidelines-for-rest-apis

christophlingg:
and we should not forget to add this brand new feature to our readme!

src/main/java/de/komoot/photon/OsmTag.java
christophlingg, line 6: btw, i think this class never gets used

src/main/java/de/komoot/photon/RequestHandler.java
sdole: Very good point. In the mappings.json, it defines `osm_key` and `osm_value`. Did you really mean to use `tag_key` and `tag_value`? If I don't hear back, I will change to `osm_key` and `osm_value`

src/main/java/de/komoot/photon/elasticsearch/Searcher.java
sdole, line 74: Will you consider this as a future improvement to be taken up as a separate issue? 

I did feel the same way as you do about the battery of if statements. Which is why I asked my question in a previous [comment](https://github.com/komoot/photon/issues/68#issuecomment-71639283). So, then, having just one JSON is ok?

Further, if you will consider my suggestion: currently, photon code is very tightly coupled and difficult to refactor without adding more if-then-elses. For example, in [App.java](https://github.com/komoot/photon/blob/30f626d92f7ff658a17c9372a8af1432dbe62e45/src/main/java/de/komoot/photon/App.java#L135), The Searcher, RequestHandler are concrete classes tied together in code. Within Searcher too, the search method does parameter replacement, query execution and then also delegates to conversion and cleansing. If these are refactored into different classes, photon code could become less bug-prone and we may be able to avoid these if-then-elses.

Of course, I don't mean to review your code, it is a very useful project as it is. I just wanted to offer my thoughts on the same lines of making the code/json less error prone.

Let me know if you would consider doing all this in a different issue as an improvement. 

Thanks!

sdole:
I made the following changes to my fork for this pull request:
1. changed tagKey and tagValue url params to osm_key and osm_value 
2. changed the osm_value field indexing from "analyzed" to "not_analyzed"
3. added a couple lines to README to describe this new feature.

Thanks!

christophlingg:
Thanks @sdole! At the moment I am preparing new data set as the 0.2.0 version had some problems. I will deploy your new feature along with the data fixes probably this week. 

I will also get my hands on the template engine stuff and check if we can have a better solution. And don't worry about your comments about the existing code. Software is never finished or perfect and good input is what it takes to make it better.

karussell:
Instead of making it more complex with a template engine I would propose to just use the Java Query API. Which we currently don't do as we would have lost the copy&paste 'feature' to tweak the JSON query. But this is not a big problem as the generated JSON can still be printed and tweaked after it is created with the Java API.

sdole:
I updated the readme to show the correct link for nominatim supported tags and also a reference to the nominatim source that @lonvia posted previously.

christophlingg:
I won't have time to work on refactoring the query with the Java Query API this week. Let's put it on the todo list. @karussell, you are the one who will know the best how to do it, if it is easy for you to do it, feel free to make a pr ;-)

sdole:
Good decision to do java and to do it separately. Happy to help for that -
if needed.

Thank you! Looking forward to using this in production. Also, let me know
if you run into bugs with the new code.

christophlingg:
This week I won't have time to do it, I am already spending quite some time to setup the new data. Would be glad if you could do it @sdole

christophlingg:
btw. the new filter is already online http://photon.komoot.de/api?q=berlin&osm_key=boundary&osm_value=administrative

sdole:
Ok. I might. :)

PR #143, opened 2015-02-20
Fix thread safety bug in Searcher

karussell:
An important fix to make the searcher class thread safe


christophlingg:
thanks @karussell, fyi @sdole

PR #144, opened 2015-02-23
Let photon connect via transport client too

karussell:
This allows to connect to external elasticsearch nodes. Beneficial if you change something in photon and don't want to restart elasticsearch nodes. There is now a new command line parameter which can be used like:
`java -Xms200m -Xmx200m -jar target/photon-0.2.2-SNAPSHOT.jar -transport-addresses 136.243.16.13:9300[,...] -other-options example`

Of course it is required to have the `photon` index and install the plugin(s) on those nodes.

BTW: the description '(default is xy)' should be removed in another issue as it is already printed from the tool when getting the help
BTW2: I removed the default clusterName="photon_v0.2.1" because it was actually never used and was misleading IMO

### DIFF
--- a/src/main/java/de/komoot/photon/CommandLineArgs.java
+++ b/src/main/java/de/komoot/photon/CommandLineArgs.java
@@ -14,7 +14,7 @@ public class CommandLineArgs {
 	@Parameter(names = "-cluster", description = "name of elasticsearch cluster to put the server into (default is 'photon')")
 	private String cluster = "photon";
         
-        @Parameter(names = "-transport-addresses", description = "the addresses of external elasticsearch nodes the client can connect to (default is an empty string which forces an internal node to start)")
+        @Parameter(names = "-transport-addresses", description = "the comma separated addresses of external elasticsearch nodes where the client can connect to (default is an empty string which forces an internal node to start)")
 	private String transportAddresses = "";
 
 	@Parameter(names = "-nominatim-import", description = "import nominatim database into photon (this will delete previous index)")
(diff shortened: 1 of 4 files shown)


src/main/java/de/komoot/photon/CommandLineArgs.java
```diff
@@ -11,13 +11,16 @@
 
 @Data
 public class CommandLineArgs {
-	@Parameter(names = "-cluster", description = "name of elasticsearch cluster to put the server into (default: photon)")
+	@Parameter(names = "-cluster", description = "name of elasticsearch cluster to put the server into (default is 'photon')")
 	private String cluster = "photon";
+        
+        @Parameter(names = "-transport-addresses", description = "the addresses of external elasticsearch nodes the client can connect to (default is an empty string which forces an internal node to start)")
```
christophlingg: could you add that it's comma separated?

karussell: done

christophlingg:
regarding BTW2: the clusterName was used here: https://github.com/komoot/photon/blob/master/src/main/java/de/komoot/photon/elasticsearch/Server.java#L90

I thought it was used to have distinguish two es nodes on one machine and prevent them to form a cluster? Was I wrong? Do you know how we can accomplish this?

Not to forget: cool pull request!

karussell:
I meant: the clusterName was used sure, but the default 'photon_xy' was *always* overwritten in the constructor.

karussell:
So now it can be even 'private final String clusterName;'

karussell:
So now the default in the CommandLineArgs is used (photon)

christophlingg:
got it, thanks for the clarification

karussell:
This separation is especially important as ElasticSearch has to create the location bias index or is this just because of the compiling for the script? There is a huge delay before one gets the location bias results. But after an initialization phase everything is very quick..

PR #146, opened 2015-02-28
much better, flexible tag filters. change from json search templates to java api

sdole:
Here are the fixes/improvements:

1. tag filter now supports multiple ways of filtering. inclusions, exclusions by tag, key, value and combinations of those. So, in theory, one can be very very specific in filtering search criteria.
2. New code no longer uses the json template based search criteria mechanism. Instead, it uses ES Java api.
3. Code is now a little more spread out into multiple classes instead of being in private methods in a handful of classes - IMO: this improves testability of code, makes it extensible, and also makes it less prone to bugs.
4. Readme has an updated description.

To follow the path of code/instructions, look at lines 139 to 143 of [App.java](/sdole/photon/blob/master/src/main/java/de/komoot/photon/App.java#L142). The old [Searcher](/sdole/photon/blob/master/src/main/java/de/komoot/photon/elasticsearch/Searcher.java) is now deprecated, but not deleted. The replacement is [SearchRequestHandler](/sdole/photon/blob/master/src/main/java/de/komoot/photon/SearchRequestHandler.java). This in turn has "Factory" classes that instantiate the required handlers to execute searches instead of direct dependency on concrete implementations. There are three steps to a search. First is to convert the native spark web request into a PhotonRequest object. During this step, we get the opportunity to validate the incoming request and do any defaults etc. Second step is to optionally setup a [tag filter](/sdole/photon/blob/master/src/main/java/de/komoot/photon/query/TagFilterQueryBuilder.java). Finally, the last step is to execute an [Elasticsearch search](https://github.com/sdole/photon/blob/master/src/main/java/de/komoot/photon/searcher/ElasticsearchSearcher.java).

I have also added [unit tests](/sdole/photon/blob/master/src/test/java/de/komoot/photon/query/TagFilterQueryBuilderSearchTest.java).

All of the Elasticsearch API setup is in class [PhotonQueryBuilder](/sdole/photon/blob/master/src/main/java/de/komoot/photon/query/PhotonQueryBuilder.java)


karussell:
Looks like a big refactoring, nice :) !

Did you already test this in production?

Some questions:
 * What is the difference between BaseElasticsearchSearcher and BasePhotonSearcher? 
 * Did you thought about avoiding the parsing code and just use a Lucene parser? (to get even more powerful queries)
 * Also there is still this `System.out.println(query);`
 * Why did you uncomment the TagFilterQueryBuilderTest only? Is it not fully replaced or replacable?

sdole:
Yes, it is sort of a big refactor. I understand your concern about testing it in production. However, what do you mean by "production"? My production is not busy enough for a good test, so, the answer is no. However:

1. I did test from a browser ("/api") against the index file I downloaded from komoot. Do you have test cases that I can manually execute against photon? I searched using my requirements and it looked fine.
2. I brought back part of TagFilterQueryBuilderTest which you mentioned in your point 4. This now has a test comparing the java output to the erstwhile json template. So, it might be worthwhile to compare the files src/test/resources/json_queries to the file /es/query.json
3. Further, I have a test named TagFilterQueryBuilderSearchTest that runs through a few test scenarios and those seem to work fine too.
4. That System.out.println() was added to searcher when I started work to see what the final json looked like before I started work. I forgot to remove it and did not notice it because it was not being executed. I have taken it out now and if you compare the fork vs master, you will notice that the only line changed is @Deprecated at the top.
5. About the difference between BaseElasticsearchSearcher and BasePhotonSearcher: I have removed BasePhotonSearcher. It was not being used.
6. About avoiding the parsing code... I have not idea how to do that... To be honest, I am very new to elasticsearch and indexing in general, I would not know where to start... Its only because photon already exists that I am able to refactor... creating this from scratch is a whole another universe for me.

Let me know what you think about testing this. I do understand what you are asking for.

sdole:
So you know that no code was lost: there was some custom coding in the Searcher.java that I have taken out into separate classes. Here those are:
1. [StreetDupesRemover](https://github.com/sdole/photon/blob/master/src/main/java/de/komoot/photon/searcher/StreetDupesRemover.java)
2. [ConvertToGeoJson](https://github.com/sdole/photon/blob/master/src/main/java/de/komoot/photon/utils/ConvertToGeoJson.java)

karussell:
Please don't take it as critique :), it is a really nice effort as the combinating-template strategy is a lot more error prone IMO.

Regarding the Lucense-parsing: there is an [ElasticSearch query](http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html) which supports this, although this query type cannot be used as direct user input it is really nice to do simple filtering and arbitrary combinations.

sdole:
No, not taking as a critique. :)

OK, I saw the link you sent. I will read some more and try to learn. For now, this pull request simply replicates what was in the json template. That way, there are fewer changes to test. You know what I mean?

src/main/java/de/komoot/photon/query/PhotonRequestFactory.java
```diff
@@ -0,0 +1,120 @@
...
+        try {
+            limit = Integer.valueOf(webRequest.queryParams("limit"));
+        } catch (NumberFormatException e) {
+            limit = 15;
+        }
+        Point locationForBias = null;
+        try {
+            Double lon = Double.valueOf(webRequest.queryParams("lon"));
+            Double lat = Double.valueOf(webRequest.queryParams("lat"));
+            locationForBias = new GeometryFactory(new PrecisionModel(), 4326).createPoint(new Coordinate(lon, lat));
```
christophlingg: there is no need to create a geometry factory for every request, i can be a static final field

src/main/java/de/komoot/photon/searcher/AbstractPhotonRequestHandler.java
```diff
@@ -0,0 +1,40 @@
...
+
+    private ElasticsearchSearcher elasticsearchSearcher = new BaseElasticsearchSearcher();
+
+    @Override
+    public final List<JSONObject> handle(R photonRequest) {
+        TagFilterQueryBuilder queryBuilder = buildQuery(photonRequest);
+        Integer limit = photonRequest.getLimit();
+        SearchResponse results = elasticsearchSearcher.search(queryBuilder.buildQuery(), limit);
+        if (results.getHits().getTotalHits() == 0) {
+            results = elasticsearchSearcher.search(queryBuilder.withStrictMatch().buildQuery(), limit);
```
christophlingg: shouldn't it be the other way around? first a strict match query. if there is no match, make a less strict query.

src/main/java/de/komoot/photon/utils/ConvertToJson.java
```diff
@@ -0,0 +1,96 @@
...
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Convert a elasticsearch {@link SearchResponse} into a list of {@link JSONObject}s
+ * <p/>
+ * Created by Sachin Dole on 2/20/2015.
+ */
+@Slf4j
+public class ConvertToJson implements OneWayConverter<SearchResponse, List<JSONObject>> {
```
christophlingg, line 20: i did not get this one: what's the difference between ConvertToJson and ConvertGeoToJson?

christophlingg:
cool that's a big pull request with a lot of refactoring! I left some commands and maybe found a bug...

I see logic got separated which is a step forward. You are right, it makes testing much easier.

sdole:
You are right @christophlingg about the bug. I committed a fix with changes to the unit test. I also made the geometryfactory private static final. added some java docs.

The difference between ConvertToJson and ConvertToGeoJson is that ConvertToJson converts SearchHits to a list of JSON objects. ConvertToGeoJson then converts to geojson format. This follows the exact pattern of code in Searcher and RequestHandler before refactoring. The Searcher class first converts SearchHits to a json and then operates on it to remove dupes etc. Then, in the RequestHandler, there are a few more lines of code to convert into GeoJson. I wanted to keep those lines of code as is, so, I simply replicated everything - except in different classes instead of private methods.

Thanks for your words. I am glad that you and @karussell feel that this is a good refactor and a step forward. I am also glad that you feel that this makes testing much easier. 

Let me know how this looks. Looking forward to this getting into production on your website! :-)

sdole:
If there is anything else you have questions about, please ask. I will be happy to change code if you find issues or have concerns.

christophlingg:
sorry @sdole , at the moment i am quite busy. i'll make an 0.2.2 release tomorow

sdole:
Absolutely no worries. Thanks for the reply.





        Sent by Outlook for Android




On Wed, Mar 4, 2015 at 12:28 PM -0800, "Christoph Lingg" <notifications@github.com> wrote:












sorry @sdole , at the moment i am quite busy. i'll make an 0.2.2 release tomorow

PR #150, opened 2015-03-10
Avoid static client

karussell:
Not sure if this is important, but I don't like static dependencies and prefer dependency injection (even if it is 'by hand'). 

BTW: Not sure if some of the used patterns are too complex like the factories but that is probably for another pull request :)

### DIFF
--- a/src/main/java/de/komoot/photon/App.java
+++ b/src/main/java/de/komoot/photon/App.java
@@ -18,11 +18,7 @@ import static spark.Spark.*;
 
 @Slf4j
 public class App {
-    private static Client esClient= null;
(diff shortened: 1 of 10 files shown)


sdole:
I feel like those patterns, though a little complex are simply there to separate functionality into separate, easily testable, extensible units. If you allow me to use jargon, those patterns promote polymorphism and reduce coupling. But, software is never perfect, it can always be improved (to paraphrase @christophling). Yes, di is better, even if by hand.

karussell:
In complex software more complex pattern is necessary and good, but here it might make more sense to avoid abstraction but of course, just my opinion :)

Issue #145, opened 2015-02-23
Replave mvel with groovy?


karussell:
I'm fighting a bit to make the separation of photon and elasticsearch happening and one of the issues is that mvel is no longer the default scripting language for elasticsearch https://github.com/elasticsearch/elasticsearch/issues/7029

karussell:
Also it looks elasticsearch should be updated to at least 1.4.3: http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/modules-scripting.html

karussell:
For the external ES you have to install:
`plugin -install elasticsearch/elasticsearch-lang-mvel/1.4.1`

christophlingg:
are you using elasticsearch without photon there? this dep should ensure the plugin to be installed: https://github.com/komoot/photon/blob/master/pom.xml#L16

karussell:
I mean a plain 'external' installation via the debian package, for this one needs the `scripts` folder in the `config/` and this mvel plugin

Issue #151, opened 2015-03-10
Critical security issue


karussell:
We should upgrade to ElasticSearch its latest version and disable dynamic scripting in the settings. I found the following settings:

```yaml
script.disable_dynamic: true
script.groovy.sandbox.enabled: false
```

PR #152, opened 2015-03-10
Upgrade to ES 1.4.4, replaced mvel with Groovy

karussell:
This PR contains two commits where first mvel was replaced with Groovy, then ES was upgraded to the latest 1.4.4. All test passed but didn't tested it yet in production or with the nominatim tests.

This PR fixes partially #151 

Still disabling dyncamic scripting in the config is to be done as we don't need it I think?


christophlingg:
about the dynamic scripting, access to elasticsearch is only granted locally: https://github.com/komoot/photon/blob/master/src/main/java/de/komoot/photon/elasticsearch/Server.java#L71 I guess this is already sufficient in terms of security as you need access to the local machine.

however, disabling dynamic scripting is still a good option. you can hard code it for now or add a config file similar to what we already do with our scripts:
https://github.com/komoot/photon/blob/master/src/main/java/de/komoot/photon/elasticsearch/Server.java#L154

ref #151

karussell:
Well, as I use more than one ES instance and the transport client this line is not sufficient :)

> add a config file similar to what we already do with our scripts:

yes

karussell:
Also mvel seems to be stuck at 1.4.1 so it might be better to move to the default groovy

karussell:
BTW: this fixes #145

christophlingg:
ah ok, i thought you wanted to add the disabling of the dynamic scripting first. let's go

karussell:
I could, yeah but ;)

dynamic scripting is done either in some external servers (if using transport addresses) or via the config you mean in the java code.

Issue #161, opened 2015-03-30
Add suburb information


nathan-muir:
Related to #30 - In Australia addresses primarily use `Suburb` over `City` - addresses are usually in the format:

```
5 StreetName StreetType
Suburb State Postcode
```

`City` represents administrative zones that are not often used or seen.

So - please consider making the `suburb` field available in the resulting API response.

lonvia:
Adding more fields is not going to scale because we end up with dozens of different requests for local tagging quirks. The way to go here is to localize the exports on a country level, e.g. for Australian addresses put the suburb info in the city field.

christophlingg:
thanks for your feedback @nathan-muir and letting us know what Australian users expect.

we could go for @lonvia's approach but we would return suburbs in a city field which does not sound correct to me...? A new field would not lead to an additional request in photon (probably in nominatim it would). As long as data storage is not increased significantly, a new field sounds so be the best option. 

Some investigation will be necessary to check what is best.

nathan-muir:
@christophlingg An extra field sounds good. 

It does look a bit fuzzy between "city" and "suburb"  in the list of administrative levels- http://wiki.openstreetmap.org/wiki/Nominatim/Development_overview#Indexing.2Faddress_calculation

It's difficult to decide what would be a reasonable subset of those levels for searching / display.

amnesia7:
@christophlingg is this the type of thing that the [address-formatting](https://github.com/lokku/address-formatting) project was trying to resolve?

avently:
@christophlingg What do you think about including country's ISO code in search results?

christophlingg:
@avently what's your motivation?

avently:
For now, I will not use Photon. It doesn't fit my needs. You made it for large search specific projects but I need only small amount of features. So I made my own thing based on my instance of orientdb and self-extracted data from planet.osm using postgis and node.

phibo23:
I'd like a country code in the results, too. At one point I'm converting the results into our own model classes and our address format requires a country code. Mapping all the localized country strings to iso codes myself seems unnecessarily complicated from my point of view, because the codes are already present in nominatim.

christophlingg:
So far, country code was not needed. I did not include it in the index as it would store the *same* information twice in the index, having 100 million entries this would cost some disk space.

Anyway, probably it would not be that much of an impact.

maraev:
Christoph, is it possible to merge my index of Switzerland with your 31Gb
photon_data?
[...]

PR #162, opened 2015-03-31
Added `suburb` to index/ results.

nathan-muir:
related to #161 - Just decided to hack in the suburb field. Appears to work well enough.

On a small dataset - index appears to have grown by about 10% (from 550mb to 610mb)

### DIFF
--- a/es/mappings.json
+++ b/es/mappings.json
@@ -68,6 +68,17 @@
           }
         }
       },
+      "suburb": {
+        "properties": {
+          "default": {
+            "type": "string",
+            "index": "no",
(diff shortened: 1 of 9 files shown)


christophlingg:
Great, that is exactly how I would have implemented it too. I won't merge the pr right now as we need to test index growth on a global scale first.

cammanderson:
Hi Guys, 

Would love to see this feature added. Appears to create some gaps in the presentation of addresses without having 'locality/suburb' within the address. 

Any chance this could be reviewed?

Thanks
Cameron

Issue #19, opened 2014-05-01
enable reverse geocoding


christophlingg:
... and let other people know your feature on the project website

Svantulden:
I am experimenting a bit with enabling reverse geocoding for my Photon installation. Is @karussell's query above still valid for the current version of ES? I am getting a `SearchParseException` with a nested `No query registered for [sort]` exception when I `QUERY_AND_FETCH`. I do not have much experience with ES, but maybe one of you can help me?

christophlingg:
hm, we updated elasticsearch from 1.1 to 1.3.1 and dynamic scripting isn't allowed anymore. for the sake of security you need to place the script in the [script folder](https://github.com/komoot/photon/tree/master/es/scripts) and add your file [here](https://github.com/komoot/photon/blob/master/src/main/java/de/komoot/photon/elasticsearch/Server.java#L122) to be copied to the right place.

I am not totally sure that is your problem, but it would make sense.

There is also a good docu for sorting by distance btw: http://www.elasticsearch.org/guide/en/elasticsearch/guide/current/sorting-by-distance.html

karussell:
@christophlingg There should not be a need for dynamic scripting for this. @Svantulden are you using some external library or directly the JSON request? If external lib make sure you use it correctly: http://stackoverflow.com/a/20175407/194609

Svantulden:
Thank you both for your very quick reply! @karussell you were indeed right that the ES library Photon uses already adds `query = { }` to the json query which resulted in the parse exception. 

I got reverse geocoding to work in my Photon installation with the following code: 

**The query:**
```JSON
{
    "filtered": {
      "query": {
        "match_all": {}
      },
      "filter": {
        "geo_distance": {
          "distance": "5km",
          "coordinate": [
            ${lon},
            ${lat}
          ]
        }
      }
    }
}
```

**The searching & sorting code:**
```JAVA
SearchResponse response = client.prepareSearch("photon").setSearchType(SearchType.QUERY_AND_FETCH)
                        .setQuery(query)
                        .addSort(SortBuilders.geoDistanceSort("coordinate").point(lat, lon).order(SortOrder.ASC))
                        .setSize(1)
                        .setTimeout(TimeValue.timeValueSeconds(7))
                        .execute()
                        .actionGet();
```

[...]

PR #164, opened 2015-04-23
Added Reverse Geocoding API & Remove old code support

Svantulden:
# Reverse Geocoding in Photon

So I finally had time to merge my old reverse geocode code for Photon with the newly refactored Photon. Sorry for the delay. This PR has basic reverse geocoding support, which lacks some features (described further on) and needs some heavy feedback, as this should be another prime feature for Photon. 

## Some info

First some info about the way I implemented reverse geocoding into Photon. This code has the following features: 

- Give a coordinate and an optional language to Photon, and it returns the closest place in the Photon index. The data gets returned in the same trusted format as forward geocoding (a GeoJSON `FeatureCollection`). I kept this pretty simplistic. It returns an empty array when no result can be found. 
- The reverse geocode code has been written for the 'old' style and the new refactored style. The old way has a different API call format from the new way:
  - Call `/api?reverse=true&lat=51.830866&lon=5.8388302&lang=nl` for a basic reverse query on the **old** way
  - Call `/reverse?lat=51.830866&lon=5.8388302&lang=nl` for a basic reverse query on the **new** way
- To switch between the new and the old way, uncomment/comment the corresponding handlers in `App.java`, just like it works in the current master
- The **old** way works with a JSON query file (`query_reverse.json`) which is basically what @karussell made in #19 . The **new** way works with the ElasticSearch Java API, like the forward geocoding does. 
- The query is based on a limit of a `5km` radius that it searches around the given coordinate and returns the best (closest) result. 
- I've tried to adhere to the new code structure and style by using factories and kind of mirroring the forward geocoding code with specific changes for this functionality. Also need some feedback here, if this fits the refactored structure.

## Production usage

We use this code in production a lot. So I can safely say that it is pretty stable. To give you an idea, we have on peak around 50 concurrent users on the geocoder (reverse and forward), with a small (5gb) index of The Netherlands, Belgium and Germany combined. This needs to be tested with a global dataset where performance/timings should be tested on also. 

I would appreciate it if someone other than me uses the global dataset to test this, because I am currently running out of storage on my small SSD to test it properly. 

## TO-DO/Missing features

- [x] Remove old code support
- [x] ~~The newly added `osm_tag` filtering is currently missing in this PR~~ (new pr)
- [x] Documentation on how to use the API (possibly in new PR?)
- [x] ~~Decide if `5km` is a nice sweet-spot for the range Photon searches in, or optionally make it configurable~~ (new pr)
- [x] Unit tests for reverse geocoding are currently missing in this PR

Thanks in advance for reviewing this PR and like I said earlier, **all feedback is welcome**.


karussell:
Wow, nice work!

src/main/java/de/komoot/photon/ReverseSearchRequestHandler.java
```diff
@@ -0,0 +1,59 @@
...
+        this.requestHandlerFactory = new ReverseRequestHandlerFactory(new ReverseElasticsearchSearcher(esNodeClient));
+    }
+
+    @Override
+    public String handle(Request request, Response response) {
+        R photonRequest = null;
+        try {
+            photonRequest = reverseRequestFactory.create(request);
+        } catch (BadRequestException e) {
+            halt(e.getHttpStatus(), "bad request: " + e.getMessage());
```
karussell, line 44: here a return is missing (?)

src/main/java/de/komoot/photon/query/ReverseRequestFactory.java
```diff
@@ -0,0 +1,44 @@
...
+        language = language == null ? "en" : language;
+        languageChecker.apply(language);
+        
+        Point location = null;
+        try {
+            Double lon = Double.valueOf(webRequest.queryParams("lon"));
+            Double lat = Double.valueOf(webRequest.queryParams("lat"));
+            location = geometryFactory.createPoint(new Coordinate(lon, lat));
+        } catch (Exception nfe) {
+            //ignore
```
karussell: why ignore and then check for 'location == null'?

src/main/java/de/komoot/photon/ReverseSearchRequestHandler.java
Svantulden, line 44: I've taken this from the existing `SearchRequestHandler.java` that is being used in Photon right now. I think there is no return there because it is redundant. Spark's `halt()` already throws a `HaltException` which breaks out of the `handle()` method. Correct me if I'm wrong though!

src/main/java/de/komoot/photon/query/ReverseRequestFactory.java
Svantulden: Good point! I'll change this

Svantulden:
Thanks for the feedback!

src/main/java/de/komoot/photon/ReverseSearchRequestHandler.java
karussell, line 44: ups, then it is perfect of course :)

Svantulden:
Any thoughts about this approach @christophlingg @yohanboniface ?

karussell:
For me it looks good but I didn't had the time to test this in production yet.

src/main/java/de/komoot/photon/PhotonDoc.java
```diff
@@ -81,8 +81,6 @@ public boolean isUsefulForIndex() {
 
 		if(name.isEmpty()) return false;
 
-		if(linkedPlaceId > 0) return false;
-
-		return true;
+		return linkedPlaceId <= 0;
```
christophlingg: intellij is continuously recommending to simplify this code this way, but i think it is very hard to understand this way

src/main/java/de/komoot/photon/RequestHandler.java
```diff
@@ -29,45 +29,55 @@ protected RequestHandler(String path, Searcher searcher, String languages) {
...
-
-        List<JSONObject> results = searcher.search(query, lang, lon, lat, osmKey,osmValue,limit, true);
-		if(results.isEmpty()) {
-			// try again, but less restrictive
-			results = searcher.search(query, lang, lon, lat, osmKey,osmValue,limit, false);
-		}
+                String osmKey = request.queryParams("osm_key");
+                String osmValue = request.queryParams("osm_value");
+                List<JSONObject> results;
+                if (reverse != null && reverse.equalsIgnoreCase("true")) {
```
christophlingg: is the true check really needed?

christophlingg:
Great work @Svantulden !

I am not super happy to have both solution (json files and ES API) at the same time. Shouldn't we better go for only one solution (the best one) and avoid duplicate logic that we need to test, maintain and improve twice?

Svantulden:
Thanks for the feedback @christophlingg !

>I am not super happy to have both solution (json files and ES API) at the same time. Shouldn't we better go for only one solution (the best one) and avoid duplicate logic that we need to test, maintain and improve twice?

I totally agree. The reason why I made both versions is because currently Photon supports both old and new ways of forward geocoding. As there wasn't much communication about the refactoring (as far as I know of), I thought I'll supply both, just as forward geocoding has at the moment. 

Also this wasn't more effort for me, as I wanted to move my already written reverse geocoding code to the new structure anyways. 

Can you tell me what direction Photon is going in? Is it moving fully to the ES api and drop the 'old' support? If so, I'll change this PR to only support the ES api way, no problem.

src/main/java/de/komoot/photon/RequestHandler.java
Svantulden: Not really. Good point!

src/main/java/de/komoot/photon/PhotonDoc.java
Svantulden: Yeah I agree. Sometimes IntelliJ helps me with some proper structure, but in this case it indeed makes it harder to understand.

christophlingg:
yes drop the support for the old version!

the refactoring happend when @sdole included the tag filters see https://github.com/komoot/photon/pull/146 there are some information about his concept.

I am not aware why he left the old code, probably it can be deleted too as it is not used any more.

sdole:
@svantulden, I left the 'old' json as it was after refactoring for posterity and later removal. I will submit a pull request which will delete all unused files. Great work on your PR! Thanks.

sdole:
You can also go ahead and delete the old code. I just saw @christophlingg's suggestion.

Svantulden:
Nice, thanks guys! I've now removed the old unused code. Please tell me if I missed something. 

Now the approach is clear, I'll start work on implementing the `osm_tag` filtering for the reverse geocoding and add a few tests.

sdole:
Let me know if you have questions. Thanks.

christophlingg:
we should not forget to mention it on the readme page!

christophlingg:
@Svantulden reverse geocoding is already a great feature without having tag filters. We merge this pull request already to master and leave the filters for a future improvement, what do you think?

Svantulden:
Sorry for the massive radio silence on my part! I was struggling with adding osm_tag filters and after that I didn't have much time to work on it. 

Good idea to seperate them to allow users to use this feature now, albeit without osm_tag filtering. 

I'll whip up some documentation and some basic tests to prepare for merging.

christophlingg:
great, the smaller the steps, the more fun it is!

Svantulden:
I've added small and basic tests for this feature. If you could review this, we may be ready to merge!

christophlingg:
I will have a look on it, hopefully today!

christophlingg:
great @Svantulden , the new version is now online! reverse geocoding is working great!

http://photon.komoot.de/reverse/?lat=45&lon=12.1&lang=fr

Svantulden:
Amazing, thanks for merging!

stephan75:
Great! Is this new feature worth mentioning in and osm news channels like wochennotitz vor weeklyosm.EU? Is there and blog article about that yet?

christophlingg:
@stephan75 i think it is worth mentioning it but we do not have a blog yet ;-(

PR #167, opened 2015-05-15
Allow either protocol for website display

amnesia7:
To allow product/demo website to be output as http or https


christophlingg:
it works like a charm: https://photon.komoot.de/

very cool @amnesia7

amnesia7:
Awesome!

PR #180, opened 2015-07-27
Close server and client properly, fixes #163

### LINKED ISSUE #163: Photon doesn’t quit after import/dump actions
Currently, with Photon from commit 1abeb523, running the command below

    java -jar /photon/target/photon-0.2.3-SNAPSHOT.jar -nominatim-import -host localhost -port 5432 -database nominatim -user root -languages de

the following message is printed to the screen but Photon does not terminate.

    […]
    [main] INFO de.komoot.photon.nominatim.NominatimConnector - finished import of 927,553 photon documents.
    [main] INFO de.komoot.photon.App - imported data from nominatim to photon with languages: de

Reading issue #64, Photon should actually stop and reading Photon’s source code in file `App.java`, it should actually return.

                if(args.isNominatimImport()) {
                        startNominatimImport(args, esServer, esClient);
                        return;
                }

Could there be a problem with closing the connection to the database server or some other process?

karussell:
Tested via `java -jar target/photon-0.2.3-SNAPSHOT.jar -recreate-index` and the process now always exits

### DIFF
--- a/src/main/java/de/komoot/photon/App.java
+++ b/src/main/java/de/komoot/photon/App.java
@@ -43,20 +43,25 @@ public class App {
 		}
 
 		final Server esServer = new Server(args).start();
-		Client esClient = esServer.getClient();
+                try { 
+                    Client esClient = esServer.getClient();                                
 
-		if(args.isRecreateIndex()) {
-			startRecreatingIndex(esServer);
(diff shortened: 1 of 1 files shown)


christophlingg:
thanks @karussell

PR #188, opened 2015-09-16
Only shutdown if not starting API

### LINKED ISSUE #163: Photon doesn’t quit after import/dump actions

karussell:
Should finally fix #163

### DIFF
--- a/src/main/java/de/komoot/photon/App.java
+++ b/src/main/java/de/komoot/photon/App.java
@@ -42,17 +42,20 @@ public class App {
 			return;
 		}
 
+                boolean shutdownES = false;
 		final Server esServer = new Server(args).start();
                 try { 
                     Client esClient = esServer.getClient();                                
 
                     if(args.isRecreateIndex()) {
+                            shutdownES = true;
                             startRecreatingIndex(esServer);                        
                             return;
                     }
 
                     if(args.isNominatimImport()) {
-                            startNominatimImport(args, esServer, esClient);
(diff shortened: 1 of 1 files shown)


christophlingg:
what about writing `esServer.shutdown();` at the end of both if clauses? would save the try-catch  and the extra variable?

karussell:
the try-finally clause makes shutting down ES even if an exception appear ... so under any circumstances. I think this is safer, although not that clean.

karussell:
Ping :)
This is an important fix and I stumbled over this issue again (after merging the recent PR).

Let me know how you want this change!

christophlingg:
sorry, i lost track of this pr

PR #218, opened 2016-02-17
support for arbitrary address import (csv files)

christophmayrhofer:
add support for:
csv file import (supported columns: lat,lon,name,housenumber,street,postcode,city,state,countrycode,importance,osmType,tagKey,tagValue)

support to specify "all" as language paramenter

new "append" parameter to add imported data to an existing index without deleting the old one

### DIFF
--- a/src/main/java/de/komoot/photon/App.java
+++ b/src/main/java/de/komoot/photon/App.java
@@ -5,6 +5,8 @@ import com.beust.jcommander.ParameterException;
 import de.komoot.photon.elasticsearch.Server;
 import de.komoot.photon.nominatim.NominatimConnector;
 import de.komoot.photon.nominatim.NominatimUpdater;
+import de.komoot.photon.csv.CsvConnector;
+import com.neovisionaries.i18n.LanguageCode;
 import lombok.extern.slf4j.Slf4j;
 import org.elasticsearch.client.Client;
 import spark.Request;
@@ -29,6 +31,16 @@ public class App {
 			log.warn("could not start photon: " + e.getMessage());
 			jCommander.usage();
 			return;
+		}
+                
+                if(args.getLanguages().equals("all")) {
+                    java.util.ArrayList<String> languageList = new java.util.ArrayList<String>();
+                    for (LanguageCode lang: LanguageCode.values())
+                    {
+                        languageList.add(lang.name());
+                    }
+                    String[] languages = languageList.toArray(new String[languageList.size()]);
+                    args.setLanguages(String.join(",", languages));
 		}
 
 		// show help
@@ -56,6 +68,12 @@ public class App {
 			if(args.isNominatimImport()) {
 				shutdownES = true;
 				startNominatimImport(args, esServer, esClient);
+				return;
+			}
+                        
+                        if(args.isCsvImport()) {
+				shutdownES = true;
+				startCsvImport(args, esServer, esClient);
 				return;
 			}
 
@@ -109,12 +127,14 @@ public class App {
 	 * @param esNodeClient
 	 */
 	private static void startNominatimImport(CommandLineArgs args, Server esServer, Client esNodeClient) {
-		try {
-			esServer.recreateIndex(); // dump previous data
-		} catch(IOException e) {
-			log.error("cannot setup index, elastic search config files not readable", e);
-			return;
-		}
+                if(!args.isAppendImport()){
+                    try {
+                            esServer.recreateIndex(); // dump previous data
+                    } catch(IOException e) {
+                            log.error("cannot setup index, elastic search config files not readable", e);
+                            return;
+                    }
+                }
 
 		log.info("starting import from nominatim to photon with languages: " + args.getLanguages());
 		de.komoot.photon.elasticsearch.Importer importer = new de.komoot.photon.elasticsearch.Importer(esNodeClient, args.getLanguages());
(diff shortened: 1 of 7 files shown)


karussell:
Thanks, looks good. Why was a modification to `add(PhotonDoc doc)` necessary? And would you mind adding a test for this import?

christophmayrhofer:
The previous implementation of add(PhotonDoc doc) required a unique placeId (as it is the case for OSM documents). The addresses for csv import however, may not have a UID and I therefore allowed for not setting an explicit ID (ES will automatically generate one if it is not provided).

Regarding tests... I am actually a node.js developer and this is my first try on java (since I really needed that feature :) I will try to read up on how to do the tests properly in java once I can free up some extra time (probably won't happen within the next 10 days).

src/main/java/de/komoot/photon/App.java
```diff
@@ -30,6 +32,16 @@ public static void main(String[] rawArgs) {
...
+                
+                if(args.getLanguages().equals("all")) {
+                    java.util.ArrayList<String> languageList = new java.util.ArrayList<String>();
+                    for (LanguageCode lang: LanguageCode.values())
+                    {
+                        languageList.add(lang.name());
+                    }
+                    String[] languages = languageList.toArray(new String[languageList.size()]);
+                    args.setLanguages(String.join(",", languages));
+		}
```
christophlingg, line 44: allowing all languages may have unwanted side effects: every language needs some extra storage, having hundreds of languages may multiply the index by 10 (I haven't done the exact math). this parameter promotes wrong usage, it should be mentioned in the README and a warn message is appropriate: 'log.warn("...")'

christophlingg, line 44: are you sure you want that feature?

karussell, line 44: Maybe move that feature out in another issue?

src/main/java/de/komoot/photon/PhotonDoc.java
```diff
@@ -75,6 +75,9 @@ public static PhotonDoc create(long placeId, String osmType, long osmId, Map<Str
 	}
 
 	public boolean isUsefulForIndex() {
+            
+                if(centroid == null) return false;
+
```
christophlingg, line 80: why was it necessary?

src/main/java/de/komoot/photon/csv/CsvConnector.java
```diff
@@ -0,0 +1,235 @@
...
+                    String tagValue = lineData(nextLine, "tagValue"); // set default to yes?
+                    
+                    // All other properties can be used universally for csv imports
+                    Envelope bbox = null;
+                    Point centroid = null;
+                    if(lineData(nextLine, "lon") != null && !lineData(nextLine, "lon").isEmpty()
+                       && lineData(nextLine, "lat") != null && !lineData(nextLine, "lat").isEmpty()){
+                        double lon = Double.parseDouble(lineData(nextLine, "lon"));
+                        double lat = Double.parseDouble(lineData(nextLine, "lat"));
+                        GeometryFactory gf = new GeometryFactory();
```
christophlingg, line 135: you can create a static field for gf, no need to initialize it for every doc

src/main/java/de/komoot/photon/elasticsearch/Importer.java
```diff
@@ -34,8 +35,14 @@ public Importer(Client esClient, String languages) {
...
 		try {
-			this.bulkRequest.add(this.esClient.prepareIndex(indexName, indexType).
-					setSource(Utils.convert(doc, languages)).setId(String.valueOf(doc.getPlaceId())));
+                    XContentBuilder source = Utils.convert(doc, languages);
+                    long id = doc.getPlaceId();
+                    if(id > 0){
+                        this.bulkRequest.add(this.esClient.prepareIndex(indexName, indexType).setSource(source).setId(String.valueOf(id)));
+                    }
+                    else{
+                        this.bulkRequest.add(this.esClient.prepareIndex(indexName, indexType).setSource(source));
```
christophlingg, line 44: that's not too nice: first you require the implicit knowledge that an id less than 1 is not a real id. second it may happen that an automatically created id will be overwritten by a nominatim id.

maybe it's just best to never set the id explicitly but let es do the job. we will loose the connection between a photon doc and a nominatim row, but this can be reestablished by using osm id and osm type (way, node, relation).

christophlingg:
cool, stuff. I left some notes while reviewing it.

src/main/java/de/komoot/photon/PhotonDoc.java
christophmayrhofer, line 80: OSM data always has a geometry, but I realized that many of the csv datasets that I had to deal with had some entries that were lacking the geometry information. I therefore added this check to make sure that those documents do not get added to the index (since they are not really useful for geocoding if they don't have the lat lon values).

christophlingg, line 80: ok

src/main/java/de/komoot/photon/App.java
christophlingg, line 44: @christophmayrhofer could you please update the PR and remove this option?

karussell:
There was no response and several changes since. @christophmayrhofer please update to the master and address the comments to create a new PR. Closing for now here.

PR #217, opened 2016-02-10
Update to ES 2.2

masda70:
ES 2.2

Tests are passing, but I could not test Nominatim import and have not yet tested with some real data.
- References to FilterBuilder(s) changed to QueryBuilder(s) (Breaking API change in ES 2.0). Note that since `and`/`or` queries are now implemented via `bool`, some methods are now deprecated and should be updated before ES 3.0 (https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking_20_query_dsl_changes.html).
- Updated `elasticsearch-wordending-tokenfilter-filter` to ES 2.2 (repo available at https://github.com/masda70/elasticsearch-wordending-tokenfilter)
- Groovy module installation now required for local node.
- Cleaned-up Postgis dependency (now only uses postgis-jts), otherwise ES throws a JarHell exception https://discuss.elastic.co/t/elasticsearch-2-0-java-api-jarhell-problems/34776
- Test gateway type is not NONE anymore (None throws an error).
- Removed `_id.path: id` from mapping.json (already deprecated in 1.x).
- Added `delete-by-query` plugin, which is needed for tests (https://www.elastic.co/guide/en/elasticsearch/plugins/2.2/plugins-delete-by-query.html)
- `name.similarity` and `collector.similarity` in `mappings.json` were throwing errors. Apparently, they were not properly defined (not doing anything). A similarity is now only defined for `collector.properties.default.similarity`, although I'm not sure if this want is needed.

PR #219, opened 2016-02-18
Reverse geocoding performance improvements and Radius/Limit parameters

masda70:
Includes everything from this pull https://github.com/komoot/photon/pull/217
Plus : 
- similarity is by default BM25 for the whole index (coherence).
- reverse queries take advantage of ES2.2 geo distance query performance improvements (actually nothing extra is required, simply cleaned-up the query builder from unnecessary calls. 
- reverse query now take radius and limit parameters. Default reverse geocoding lookup radius set to 1km (previously 5km), limit 1.
# Benchmarks

Java 8GB RAM, full globe index (160M documents).
**Index size**
- ES 2.2: 90GB index
- ES 1.7: 64GB index

**Requests per second**
- ES 2.2 (this pull), Spinning disk: 25 to 65 requests / second, depending on data locality.
- ES 2.2 (this pull), SSD: TO BE TESTED.
- ES 1.7, SSD: 1 request/second.
- ES 1.7, SSD, indexed lat_lon : 50-65 requests / second.

ES 2.2 with SSD test coming soon.  Expecting something like 100 requests / seconds.


christophlingg:
Hi @masda70 ,

great improvements, that's a significant leap forward for photon!

How would you like to progress now? Some changes seem to be uncritical, like the pimped reverse geocoding. But updating elasticsearch's version or changes in the mapping are more critical and might cause some problems.

In the previous PR you mentioned you were indexing with the new version already, is there a chance to check the results of your updated version with the one currently available on photon.komoot.de. If results and results order are still the same, we can be 100% sure nothing broke.

masda70:
Passing radius/limit parameters should also help improve reverse geocoding performance depending on the client requirements. We should go further and exploit the limit parameter in more sophisticated 
to also limit the number of intermediate results the server sorts.

However, I believe that it's indeed more important to check that nothing else has broken. 
For testing the results against photon.komoot.de, I can do it. Do you have a set of queries to test on?

christophlingg:
there is a geocoder testing framework: https://github.com/geocoders/geocoder-tester

it comes with a bunch of test cases you could use directly or take for inspiration:
https://github.com/geocoders/geocoder-tester/tree/master/geocoder_tester/world

masda70:
Great. I'm doing a test run with your tool. Some basic tests are failing on the original branch. Perhaps photon requirements have changed  ?

For instance the following fails :
paris%20france

christophlingg:
i would also go for equality, otherwise you can also see if the number of tests failing hasn't changed.

masda70:
I tested my local version against photon.komoot.de on a France dataset, consisting of 8742 address queries with and without location biases. 

- 40.8% full matches (all returned features match, in order).
- 3.8% lenient full matches (all returned features match, in order, after stripping their coordinates and `osm_id`)
- 31.4% lenient head matches (the first returned feature matches, after stripping its coordinates and `osm_id`), with a match score of 63.6%.
- 24% partial matches (the rest), with a match score of 56.0%.

The match score is computed as the size of the set intersection between both results divided by the size of the set union, after features were stripped from their coordinates and `osm_id`

From manual inspection of a few of the results, the non matches (27.4%) are either:
 - The 2.2 version returned a better result
 - No version returned anything close the expected result.

christophlingg:
thanks for your insights @masda70 . Did you also compare the numbers of your local installation with the `photon.komoot.de`? You changed the mapping file and moved analyzers, so if there is a significant drop in some matching, it would be a hint that some indexing was going wrong. btw is the index disk size comparable to old one?

masda70:
I don't understand your first question. 

The index sizes are :
- ES 2.2: 90GB index
- ES 1.7: 64GB index

The only thing signficant thing I changed in the analyzer/mappings settings was the default similarity. The rest was simply fixing broken 1.7 deprecations.

One thing to try would be to reindex with the older similarity, but I have no idea how it was setup (the similarity settings from 1.7 do not validate at all on 2.2), and my impression is that the default TF/IDF similarity was being used.

christophlingg:
wow, an increase of 40% in disk size. how does it come?

Regarding my first question: well you were mentioning the matching rates of your local installation with ES 2.2. How do they compare to the matching rates of photon.komoot.de using current master?

masda70:
Ok. Got some results :

ES 2.2 vs local ES 1.7
 - 41.2% full matches.
 - 3.6% lenient full matches.
 - 31.3% lenient head matches (match score 63.9%)
 - 23.9% partial matches (match score 56.6%)

Local ES 1.7 vs photon.komoot.de
- 77.9% full matches
- 3.3% lenient full matches.
- 17% lenient head matches (match score 93.2%)
- 1.7% partial matches (match score 79.6%)

christophlingg:
77.9% -> 41.2% full matches and 40% increase of disk size. something must have broken

karussell:
Maybe we can get the unit tests fixed for 2.2 and see if there is an earlier sign of problem :) ?

masda70:
Do you want to take at look at the raw results?

masda70:
I now realize that the default mapping has a source exclusion definition.

```
"_source":{
  "excludes": ["context.*"]
}
```

This means that my reindexing process did not use same source data as used by the original index. Do you any idea how much this could affect the final result?

christophlingg:
"_source":{
      "excludes": ["context.*"]
    }

that would explains the increase in disk size

masda70:
On the contrary, in my case, disk size should decrease.
I'm not indexing from Nominatim but from the original ES 1.7 cluster. This entails that my indexing process did not ever have access to `context.*` fields at all.

masda70:
Tests running with:

`test -DforkMode=never`

masda70:
After discussing with @karussell I fixed some issues with my branch. I also removed the deletebyquery dependency and simplified the clean up code ran by tests. Note that tests no longer delete the ES test folder. There is no simple workaround to it: once tests start running, plugins are loaded into the classpath, and it becomes impossible to properly remove the ES test folder. Splitting into multiple folders causes Jar hell problems. The solution is to never delete the test folder in between tests. It might good to add a global "delete ES test folder" at the beginning of all tests, or simply, make all tests share the same ES configuration instead of recreating one for each test class... Regardless, tests must be run with the option `-DforkMode=never`, otherwise concurrency problems arise with ES.

Tests were ran with:
`mvn test -DforkMode=never`

karussell:
Thanks for this! I'll test again! 

> Splitting into multiple folders causes Jar hell problems

What do you mean here? I had fixed this with the pom.xml change (see the other issue)

> The solution is to never delete the test folder in between tests. 

In my solution I just started ES once and cleared the indices instead of starting&stopping ES. But as I said: I didn't really made this running, although I think we should still avoid start&stop too often (maybe just for one test class? -> via `@BeforeClass public static void startES()`)

masda70:
> What do you mean here? I had fixed this with the pom.xml change (see the other issue)

I tried a new approach for test classes which created a random test folder for each class.This does not work since the same plugin will then be loaded into the classpath, causing JarHell. Which other issue are you exactly referring to?

> In my solution I just started ES once and cleared the indices instead of starting&stopping ES. But as I said: I didn't really made this running, although I think we should still avoid start&stop too often (maybe just for one test class? -> via @BeforeClass public static void startES())

Yes this is the better approach IMO. Otherwise Server.start() gets tested too many times in an uncontrolled fashion. Index recreate is simply the only clean-up required between each test (deletebyquery is useless). Ideally there should be a global object PhotonTestServer for which all Server tests depend on, and let this object clean itself up at the end.

karussell:
I meant this pom change: https://github.com/komoot/photon/pull/217#issuecomment-186659610

Without this change (which we should still apply) I got jar hell issues 

> deletebyquery is useless

did you get rid of it?

> The only problem is always making sure ES tests are not run concurrently.

The even could if we create random index names :)

masda70:
OK.

> did you get rid of it?

I did.

> We even could, if we create random index names :)

Yes, plus this way we could make the index name configurable.

masda70:
By the way, here is the tool I used to compare 1.7 vs 2.2 results.

https://github.com/masda70/geocoder-comparison

masda70:
Also, please not that I did not apply the commits on the branch https://github.com/komoot/photon/pull/217

I should have.

karussell:
Thanks a lot, using this PR here is indeed much better!

Still I'm getting:
```
  testFilterWithoutTagTourismAttraction(de.komoot.photon.query.TagFilterQueryBuilderSearchTest): already exists
  testValueInformationButKeyNotAmenity(de.komoot.photon.query.TagFilterQueryBuilderSearchTest): already exists
  testKeyTourismButValueNotInformation(de.komoot.photon.query.TagFilterQueryBuilderSearchTest): already exists
```

I'll see how to fix these ... maybe with the global object ...

masda70:
Great! It's nice to see things progressing. 
The global object seems the most reasonable solution. I haven't checked, but perhaps it's a good idea to check how other projects that use ES are doing tests?

reuschling:
Is there any progress on this task? Now we have June 2017, and there are again huge enhancements regarding [geocoding in elasticsearch](https://www.elastic.co/blog/elasticsearch-5-0-0-released#data-structures), in Elasticsearch 5.x, released ~one year ago.

"Lucene 6 brought a new Points datastructure for numeric and geo-point fields called Block K-D trees, which has revolutionised how numeric values are indexed and searched. In our benchmark, Points are 36% faster at query time, 71% faster at index time and used 66% less disk and 85% less memory (see Searching numb3rs in 5.0). The addition of Points means that the ip field can now support IPv6 as well as IPv4 addresses. On top of that, we have changed geo_point fields to use Lucene’s new LatLonPoint, which doubles geo-point query performance."

masda70:
I haven't touched the project for a quite a bit now. One issue is that I hadn't setup a Nominatim DB so I couldn't verify if geocoding results were the same as before this patch. If you have such DB at your disposal, maybe we can work together to finish this task.

reuschling:
Hi, setting up the nominatim DB is not such easy, so I also decided to reindex the elasticsearch index to the new version for now. The process ist still ongoing. I managed it to get all the unit tests running by deleting the index after each test method.
But  the nominatim DB is also from interesst for us: We want to make high performance surface lookups for runners GPS traces,similar to what Komoot also has in their app. But photon only has the highway attribute values in the provided dump, lacking the surface and tracktype attributes from OSM. I assumed that Komoot eats their own food for generating surface histograms, also using photon?

masda70:
Reindexing won't properly work AFAIK, unfortunately. Please read the discussion above about `"excludes": ["context.*"]`.

Also, if you ever integrate my changes in the final version, I would be interested in changing the author id I used for my commits just before you prepare a new pull request (https://help.github.com/articles/changing-author-info/). I realized a little late that I used the wrong handle.

reuschling:
I read the discussion about  the excludes, but I interpreted it a bit dfferent. I assumed that these excludes happened when the  origin 1.7  index was created out of the nominatim DB. So this data isn't in the photon index anyway. All other stuff that is relevent for searching and returning the metadata should be in the index as I seen it  in the source code. If an attribute is in the source (in the ES index source on a document), it should be also in the reindexed index - and I didn't see that one attribute was excluded from the source.

masda70:
Reindexing requires that the `_source` field is left untouched (see https://github.com/elastic/elasticsearch/pull/10915 for instance).

How are you reindexing?

reuschling:
ah - yes you are right, the context.* are not in _source. Unfortunately, these can not be returned and the context in the mapping is not indexed for searching anyway:
"context": {
        "properties": {
          "default": {
            "type": "string",
            "index": "no",
            "copy_to": [
              "collector.default"
            ]
          }
        }
      }
Further, it is copied into collector.default, which is in _source.
I re-checked the source code for occurences for context, just occurs during the mapping generation. Further, All queries build by the querybuilders don't use a context field at all. Thus, this information should be only relevant inside collector.default, which is in _source, which is reindexed. I reindex with [elasticsearch-reindex](https://github.com/garbin/elasticsearch-reindex)
Regarding your author ID: how do we handle this? Is it enough that you change it and I update my repo?

masda70:
> Regarding your author ID: how do we handle this? Is it enough that you change it and I update my repo?

If you've already commited new things into your repository with my changes included, please push somewhere and I will push a new (nuked) version with the author id changes. Then you can continue working from this new version, ditching your old. If you prefer, we can wait until you changes are ready.

If you haven't commited anything new yet, I can do the changes in my repo, and you can simply clone it.

reuschling:
Regarding the test whether the results of photon keep identical or not - it is not so simple, because the default similarity measure, as you also said masda, changed in Lucene and thus also in ES. They changed it because it is the state of the art and gives in average better results. Thus, I assume slightly different results, but these are maybe better.
For a correct check, we would have to check the quality of the results, if they are better or worser with the new similarity measure. For this, we would need a ground truth, and generating precission and recall measures. A bit challenging.
I also would assume that photon only did subjective tests in the past (plesae correct me if I'm wrong). Thus, by getting a subjective impression of the quality of the results would be -at minimum- as good as the results in the past, right?

masda70:
I agree. The suggested geocoder tester seems to be a good start as ground truth: https://github.com/geocoders/geocoder-tester

reuschling:
ah ok - but someone has to annotate a bag (at best several hundred) of query-result pairs manually in order to make this happen.
I haven't commited something yet, the changes were quite simple. Am I right that  I have to fetch both #219 and #217 ,merging them together in order to get all off your changes? The first sentence 'Includes everything from this pull #217' is wrong, right?

karussell:
We used integration tests from @yohanboniface. @christophlingg do you know what I mean :) ?

masda70:
> I haven't commited something yet, the changes were quite simple. Am I right that I have to fetch both #219 and #217 ,merging them together in order to get all off your changes? The first sentence 'Includes everything from this pull #217' is wrong, right?

So it was simpler than I expected. I rewrote the author id in my fork. As for your question : #219 includes everything from #217, so you can just pull #219. Alternatively, you can pull the following branch:  https://github.com/masda70/photon/tree/feature/multiple_reverse_results

reuschling:
hmm - you commented above: 'Also, please not that I did not apply the commits on the branch pulled by #217' (1 Mar 2016). Is this obsolete?

masda70:
I was talking about the jar hell fixes. They were not applied to the #217 pull. In any case, #217 is obsolete, and #219 has all the new features and improvements.

reuschling:
ok. Thanks a lot

reuschling:
@masda70, how did you migrate the wordending-plugin? I realized that the index migration tool I used totally failed silently in order to copy the index settings and mappings.

reuschling:
hmm - I'm not sure if the wordending plugin is necessary for some reason in current ES versions. Do we really need it? NGram-autocompletion is also possible without an extra plugin: [Edge NGram Tokenizer ES 5.4.1](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-edgengram-tokenizer.html)

masda70:
I don't think it's needed anymore. 
As for your initial question : https://github.com/masda70/elasticsearch-wordending-tokenfilter

reuschling:
hmm - migration of the mapping is pretty hard, still on ES 2.2. I now recognized that they changed the analysis syntax - and further recognized that the search_ngram analyzer that should be used for searching was never used in the 1.7 version of proton. Thus, this functionality never worked from my piont of view. Or does anybody see what I doesn't see?

 [mapping definition from git](https://github.com/komoot/photon/blob/master/es/mappings.json) 

[analyzer declaration in index settings](https://github.com/komoot/photon/blob/master/es/index_settings.json)

reuschling:
I now made an according pull request for the ES 5 migration #254

karussell:
Also closeable due to #254

Issue #249, opened 2017-06-07
Update to ES 5.x?


reuschling:
Hi, I was wondering if it is planned to update photon to a current elasticsearch version. In the 5.x line (released ~one year ago), there were again huge performance optimizations regarding [geopoint search](https://www.elastic.co/blog/elasticsearch-5-0-0-released#data-structures).

"Lucene 6 brought a new Points datastructure for numeric and geo-point fields called Block K-D trees, which has revolutionised how numeric values are indexed and searched. In our benchmark, Points are 36% faster at query time, 71% faster at index time and used 66% less disk and 85% less memory (see Searching numb3rs in 5.0). The addition of Points means that the ip field can now support IPv6 as well as IPv4 addresses. On top of that, we have changed geo_point fields to use Lucene’s new LatLonPoint, which doubles geo-point query performance."

The fact that these are numbers against ES 2.x, and photon is currently at ES 1.7, lets assume that we would really have a huge benefit, maybe factor 20, from this, as reverse lookup is sadly not really fast in photon currently.

karussell:
There is an outstanding pull request that needs a bit work to update to 2.x, maybe you can further test and polish this or even use this to do a bigger upgrade to 5.x?

reuschling:
hm ok, maybe I can arrange it to find the time - what is left on this pull request?

karussell:
See my comments in it #217 and also see #219

reuschling:
I now made an according pull request #254. Further discussions better there.

Issue #253, opened 2017-07-03
Problem with area=yes or highway=pedestrian?


karussell:
When I search for http://photon.komoot.de/api?q=Recklinghausen%20markt&osm_tag=highway&lang=de I do not get anything relevant. But for one specific amenity [a result is returned](http://photon.komoot.de/api?q=Recklinghausen%20Markt%2011%20Orthop%C3%A4dicum), the street name is missing here.

Now the most confusing thing is when I leave out the word 'Orthopädicum' I [do no longer](http://photon.komoot.de/api?q=Recklinghausen%20Markt%2011) get this result.

lonvia:
The [pedestrian way](http://www.openstreetmap.org/way/31509724) in question does not have a name. Adding only addr:street to a POI is not sufficient as Nominatim takes the name from the attached highway object. You can see that quite well in the [details for the Orthopädicum](http://nominatim.openstreetmap.org/details.php?osmtype=N&osmid=4703375785) in Nominatim. The street is listed as 'No Name' there.

karussell:
Ah, thanks - makes sense. Why is "addr:street" not taken if no highway name is present? Or is this just the contract that the POIs are only connected to each other if the street name is identical?

lonvia:
The information was not really retained until recently. Nominatim now saves the information (see https://github.com/openstreetmap/Nominatim/pull/700), so photon could now in theory use the addr:* tags where available.

lonvia:
This was implemented with #303.

Issue #198, opened 2015-10-15
Location Preference does not appear to be honored


dan-castle-rock:
I am evaluating photon for use as an open source alternative and am not seeing the results I would expect with location biasing for a geocode request. The URL is passing in a latitude/longitude centered on Portland Oregon USA. When I enter a query such as 'bois', I would expect the results to include Boise, Idaho (which is nearby) to appear in the search results, but instead am receiving results from around the globe. Being new to this library, I could be doing something incorrectly.

The URL I am passing in is: http://photon.komoot.de/api/?q=bois&lat=45.52&lon=-122.6819

christophlingg:
Our location bias fosters results nearby but does not skip results that are far away. Exact matches (in this case with the french city *Bois*) are more important than partial matches (*Boise*). In the end the exact match overweights the benefit of the location bias and the french city appears first.

if you type the entire name it works: http://photon.komoot.de/api/?q=boise&lat=45.52&lon=-122.6819

karussell:
I find this problematic too. It might related to #193

dan-castle-rock:
Thank you for the reply. I agree with Peter. For our application, location
biasing is more important than exact matching (as I suspect is true with a
lot of applications). As suggested in issue 193, are there any plans to add
a filter that would give more weight to location than exact name?

On Fri, Oct 16, 2015 at 1:44 AM, Peter <notifications@github.com> wrote:

> I find this problematic too. It might related to #193
> <https://github.com/komoot/photon/issues/193>
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/komoot/photon/issues/198#issuecomment-148656662>.
>



-- 
----------------------
Dan Nichols
Mobile Software Developer
Castle Rock Associates

christophlingg:
Different uses cases require different location biases. I think it would make sense to have

 - a location bias that is controlled by an additional request parameter
 - an additional endpoint that returns entries only within a given radius. That's cool if you want to search for stuff that is nearby / in my city

karussell:
maybe a new parameter like `distance_filter=5km` which directly filters instead of scores the results, would be sufficient and also usable in other situations.

amnesia7:
Or `proximity`. 
Maybe it could apply extra weighting to those within the proximity requested but still allow global ones.

amnesia7:
[...]

PR #254, opened 2017-07-05
Migration to Elasticsearch 5

reuschling:
This is the current result for Elasticsearch 5 migration, as asked for at #249 (from myself).

In this pull request, everything from #219 is included plus some commits for bugfixing (tests are running now at least on my machine, plus the jarhell fixes from #217. This is commit 538ee4d.

Further, I implemented the migration to ES 5. The tests are running, I also have a running server with results that looks subjectively good as a first impression (further testing needed of course). I just reindexed the old ES index into a new one for testing - currently I have no Nominatim server for testing the import code. Regarding the Nominatim import, there was also a regarding commit these days - I didn't rebased it yet for simplicity during the merge.

I am looking forwad to feedback and comments :)


simonpoole:
Maybe komoot @christophling could indicate if that have any interest in further improvements or if a fork or a move of the repo to new maintainers would be preferred.

karussell:
@simonpoole this is not necessary IMO as @lonvia has access.

@lonvia provids some help (and we the server) in the background to let @reuschling test this PR for bigger data sets, if this works we'll try this for the world wide export. Maybe we should better communicate this activity though.

lonvia:
I just had a very quick look through the change. There seem to be a lot of code style changes intermixed with the actual changes. Any chance you could get rid of the style changes? They make it hard to review now and make it harder later to see where code changes come from, should there be any trouble.

reuschling:
The code style changes were committed separately in order to avoid this confusion. I also written this to the according commit comment.

lonvia:
That's nice but doesn't help much with the review. Being the one who likely has to press the merge button, I would greatly appreciate if code style cleanups go into a separate PR.

reuschling:
I am not sure if I get your point. If the whitespace changes are in a separate commit, code change review is possible, which is the practice we handle it in our company and our open source projects. There were plenty of changes in the context of the migration, would be hard to roll it back to the former state and implement everything again with the former code style. And a whitespace change in front of the code changes would be same in a different pull request, isn't it. Missed I something? Or doesn't you want whitespace changes at all? I didn't want to generate extra work.

karussell:
> there were plenty of changes in the context of the migration, would be hard to roll it back to the former state and implement everything again with the former code style.

Couldn't you set the previous styles in your IDE and apply one further commit? (and then one could squash the changes)

I'm with @lonvia that this helps reviewing PRs, or in general: change as few things as possible in one PR :)

reuschling:
done - I did the according further commit. Hope this helps.

pom.xml
```diff
@@ -1,26 +1,56 @@
...
+			<groupId>org.elasticsearch.client</groupId>
+			<artifactId>transport</artifactId>
+			<!-- <version>5.4.1</version> -->
+			<version>5.5.0</version>
 		</dependency>
+		<!-- <dependency> -->
+		<!-- <groupId>org.elasticsearch.module</groupId> -->
+		<!-- <artifactId>lang-groovy</artifactId> -->
+		<!-- <version>2.2.0</version> -->
+		<!-- </dependency> -->
```
karussell: So we are no longer using the groovy plugin (?) but why are new files in es/modules/lang-groovy/ or misunderstood I something? If there is still some groovy required - couldn't we switch to `painless` too?

src/main/java/de/komoot/photon/query/PhotonRequestFactory.java
```diff
@@ -17,12 +18,22 @@
 public class PhotonRequestFactory {
     private final LanguageChecker languageChecker;
     private final static GeometryFactory geometryFactory = new GeometryFactory(new PrecisionModel(), 4326);
+    
+    protected static HashSet<String> m_hsRequestQueryParams = new HashSet<>(Arrays.asList("lang", "q", "lon", "lat", "limit", "distanceSort", "osm_tag"));
```
karussell: instead of distanceSort we should use distance_sort?

src/main/java/de/komoot/photon/query/PhotonRequestFactory.java
```diff
@@ -42,13 +53,25 @@ public PhotonRequestFactory(Set<String> supportedLanguages) {
         } catch (Exception nfe) {
             //ignore
         }
+        Boolean locationDistanceSort = true;
+        try {
+            if(webRequest.queryParams("distanceSort") == null)
+                locationDistanceSort = true;
+            else
+                locationDistanceSort = Boolean.valueOf(webRequest.queryParams("distanceSort"));
```
karussell: same under_score format for the parameter here IMO

src/main/java/de/komoot/photon/query/ReverseRequestFactory.java
```diff
@@ -14,26 +17,90 @@
...
+            }
             Double lat = Double.valueOf(webRequest.queryParams("lat"));
+            if(lat > 90.0 || lat < -90.00){
+                throw new BadRequestException(400, "invalid search term 'lat', expected number >= -90.0 and <= 90.0");
+            }
             location = geometryFactory.createPoint(new Coordinate(lon, lat));
-        } catch (Exception nfe) {
+        } catch (NumberFormatException nfe) {
+            throw new BadRequestException(400, "invalid search term 'lat' and/or 'lon': /?lat=51.5&lon=8.0");
+        } catch (NullPointerException nfe) {
```
karussell: Instead of catching NPE we should do an if statement here (?)

src/main/java/de/komoot/photon/query/ReverseRequestFactory.java
```diff
@@ -14,26 +17,90 @@
...
+            }
+            if(radius <= 0){
+                throw new BadRequestException(400, "invalid search term 'radius', expected a strictly positive number.");
+            }else{
+                // limit search radius to 5000km
+                radius = Math.min(radius, 5000d);
+            }
+        }
+        
+        String queryStringFilter = webRequest.queryParams("queryStringFilter");
```
karussell: queryStringFilter -> query_string_filter

src/main/java/de/komoot/photon/query/ReverseRequestFactory.java
```diff
@@ -14,26 +17,90 @@
...
+        
+        Boolean locationDistanceSort = true;
+        try {
+            if(webRequest.queryParams("distanceSort") == null)
+                locationDistanceSort = true;
+            else
+                locationDistanceSort = Boolean.valueOf(webRequest.queryParams("distanceSort"));
```
karussell: distance_sort

src/main/java/de/komoot/photon/utils/QueryToJson.java
```diff
@@ -16,12 +16,14 @@
...
-            return null;
-        }
+	try {
+	    BytesReference bytes = anItem.toXContent(JsonXContent.contentBuilder(), new ToXContent.MapParams(null))
+		    .bytes();
+
+	    return bytes.utf8ToString();
+	} catch (IOException e) {
+	    log.error("Unable to transform querybuilder to a json string due to an exception", e);
+	    return null;
```
karussell: I know this was already there but I think we should throw an exception here instead of silently accepting it, e.g. 
`throw new RuntimeException("Unable to transform ....", e);`

pom.xml
reuschling: You are right, I switched to painless (inline) at the end, which is the new ES default scripting language. I first tried the scripting with groovy and started to migrate this by copying also the according lib files to the modules. Nevertheless, groovy is now not required at the end, the last cosmetic finish would be to remove these lines and files accordingly. Waiting for further testing to be finally sure with this.

src/main/java/de/komoot/photon/query/PhotonRequestFactory.java
reuschling: If you care - at the end this is a matter of taste. We could also use distancesort.

src/main/java/de/komoot/photon/query/ReverseRequestFactory.java
reuschling: This is not from me ;)  But you are right.

reuschling: but...when I look on especially this piece of code between the try{} - indeed, I agree its the better choice. It yields to much clearer code here, and its more robust because many possible nullpointer sources are considered. And the error message is still adequate.

src/main/java/de/komoot/photon/query/PhotonRequestFactory.java
karussell: it has to be consistent IMO with the other params like osm_tag. So either always camelCase or always under_score IMO :) ... see https://github.com/graphhopper/graphhopper/issues/682 :)

src/main/java/de/komoot/photon/elasticsearch/Server.java
```diff
@@ -37,208 +40,237 @@
...
+	final File scriptsDirectory = new File(esDirectory, "config/scripts");
+	final File groovyDirectory = new File(esDirectory, "modules/lang-groovy");
+	final File painlessDirectory = new File(esDirectory, "modules/lang-painless");
 
-		// add all langs to the mapping
-		mappingsJSON = addLangsToMapping(mappingsJSON);
-		client.admin().indices().prepareCreate("photon").setSettings(IOUtils.toString(index_settings)).execute().actionGet();
-		client.admin().indices().preparePutMapping("photon").setType("place").setSource(mappingsJSON.toString()).execute().actionGet();
-		log.info("mapping created: " + mappingsJSON.toString());
+	for (File directory : new File[] { esDirectory, photonDirectory, pluginDirectory, scriptsDirectory,
```
hbruch: photonDirectory should be created before esDirectory

src/main/java/de/komoot/photon/elasticsearch/Server.java
```diff
@@ -37,208 +40,237 @@
...
+		groovyDirectory, painlessDirectory }) {
```
hbruch: groovyDirectory modules/lang-groovy must not be created any longer

hbruch:
@reuschling Thank you for preparing this PR! I'm really looking forward to it. Besides the minor adaptions I had to do to get it working (see comments in the PR), I encountered some strange results when trying it for a reference database. Of 112 adresses, only 7 were geocoded with a housenumber, only 3 of them exact matches. Before, 88 were geocoded with a housenumber, 80 of them exact matches.

I'm not to deep in elasticsearch, but might housenumbers and postalcodes be dropped by the lowercase tokenize as https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-lowercase-tokenizer.html suggests?

reuschling:
@hbruch thanks a lot for rechecking the code! What exactly do you mean with 'were geocoded with a housenumber'. Did you have a look into the elasticsearch index, or is it about the photon result? Can you maybe tell me the exact index attribute you mean? Yes the lowercase tokenizer drops the numbers, but it was used in the former photon versions also, and not on all fields, i.e. the field 'housenumber' is  'not_analyzed'.
Have you checked that the housenumbers are correctly on the according nominatim entries also?

es/index_settings.json
```diff
@@ -1,56 +1,67 @@
...
+		"analyzer":
+		{
+			"index_ngram":
+			{
+				"tokenizer": "edge_ngram",
+				"filter": ["lowercase"]	
+			},
+			"search_ngram":
+			{
+				"tokenizer": "lowercase"
```
hbruch: The lowercase _tokenizer_ suppresses numerics, whereas the formerly used lowercase _filter_ did not, I think

hbruch:
@reuschling The photon results did not return results with a matching housnumber attribute, the street was matched correctly, but the housenumber attribute was empty for most results. The housenumber are correct in nominatim and reverse geocoding in photon returns the address with housenumber as well. 
Currently, I'm importing germany and I will run geocoder-tester in the next days...

reuschling:
@hbruch I think you are perfectly right, thanks a lot for finding this issue. I changed the analyzer definitions accordingly, maybe you can check if the results are better now - it was only on the ngram-related analyzers, so not all attributes were related to this.

karussell:
Just to let you know: together with @lonvia we've done a world wide export, which is available via

`wget -O - http://download1.graphhopper.com/public/photon-db-171026.tar.bz2`

I was using the recent commit 11f9aadc835a55e044e0b33532a1a2542534f036 of this PR and then merged master into it to use the latest change from @lonvia.

Will make photon with this data available for testing soonish.

BTW: there is a test failure:

testHighRadius(de.komoot.photon.query.ReverseRequestFactoryTest): expected:<5.1> but was:<5.0>

lonvia:
The new dump is 51GB, while the latest one from master is only 40G. That's a 25% increase. Where does this come from?

karussell:
Good question :) - for the extracted data it is 77GB vs. now 108GB (+40

The test server is live at: http://46.4.67.134:2322/api?q=dresden

The reverse search is crazy fast now - wow, good work @reuschling ! Also from a quick, manual test the results seems to be better!

src/main/java/de/komoot/photon/query/ReverseRequestFactory.java
```diff
@@ -14,26 +17,90 @@
...
+        String queryStringFilter = webRequest.queryParams("query_string_filter");
+        
+        Boolean locationDistanceSort = true;
+        try {
+            if(webRequest.queryParams("distance_sort") == null)
+                locationDistanceSort = true;
+            else
+                locationDistanceSort = Boolean.valueOf(webRequest.queryParams("distance_sort"));
+            
+        } catch (Exception nfe) {
```
karussell: Maybe rethrow as a BadRequestException or why ignore?

es/index_settings.json
```diff
@@ -1,56 +1,68 @@
...
+		"filter":
+		{
+			"photonlength":
+			{
+				"min": "2",
+				"type": "length"	
+			}	
+		}	
+	}	
+}
```
karussell: For the few json files it would be good if the previous formatting can be kept (maybe this can be tweaked in the IDE like intelliJ or netbeans?). Otherwise it is hard to review.

es/mappings.json
```diff
@@ -1,199 +1,504 @@
...
+					{
+						"type": "string",
+						"index": "no",
+						"copy_to": ["collector.it"]	
+					}	
+				}	
+			}	
+		}	
+	}	
+}
```
karussell: Especially here in the mapping file only the changes would be interesting. To see if we can avoid some to reduce the data size like mentioned in the PR https://github.com/komoot/photon/pull/254#issuecomment-339947104

src/main/java/de/komoot/photon/App.java
```diff
@@ -19,149 +19,149 @@
...
+                    @Override
+                    public void run() {
+                        nominatimUpdater.update();
+                    }
+                };
+                nominatimUpdaterThread.start();
+                return "nominatim update started (more information in console output) ...";
+            }
+        });
+    }
```
karussell: Here is a formatting problem too

src/main/java/de/komoot/photon/query/PhotonQueryBuilder.java
```diff
@@ -1,293 +1,407 @@
...
-		filterBuilderForTopLevelFilter = FilterBuilders.orFilter(
-				FilterBuilders.missingFilter("housenumber"),
-				FilterBuilders.queryFilter(
-						QueryBuilders.matchQuery("housenumber", query).analyzer("standard")
-				),
-				FilterBuilders.existsFilter(String.format("name.%s.raw", language))
-		);
-		state = State.PLAIN;
-	}
+    private FunctionScoreQueryBuilder m_finalQueryWithoutTagFilterBuilder;
```
karussell: What does the prefix `m_` mean?

src/main/java/de/komoot/photon/query/PhotonQueryBuilder.java
```diff
@@ -1,293 +1,407 @@
...
+
+	// this is former location-biased-score, now inline
+	// exemplary debugged score: 0.5000002150154673, is multiplied with the
+	// scores from other query parts: 2150.804 (basequery) * 31,7
+	// (importance doc value-script,
+	// former general-score)
+	// the score from the location distance is very small with respect to
+	// the score of the other query parts, thus it has a very small, up to
+	// zero, influence to the
+	// final sorting of the documents
```
karussell: There seems to be a formatting problem too for this comment using too many lines

pom.xml
```diff
@@ -1,26 +1,48 @@
...
+		</dependency>
+		<dependency><!-- required by elasticsearch -->
+			<groupId>org.apache.logging.log4j</groupId>
+			<artifactId>log4j-api</artifactId>
+			<version>2.7</version>
+		</dependency>
+		<dependency><!-- required by elasticsearch -->
+			<groupId>org.apache.logging.log4j</groupId>
+			<artifactId>log4j-core</artifactId>
+			<version>2.7</version>
```
karussell, line 35: There is a problem when I start photon:
`ERROR StatusLogger No log4j2 configuration file found. Using default configuration: logging only errors to the console.`

As logging is really ugly in the current version (not sure if this is my fault or valid for the official release too) we should really make sure that logging when doing the nominatim export is good (with date&time and throws stack traces properly). But probably we should move this into a separate issue

src/main/java/de/komoot/photon/elasticsearch/Server.java
```diff
@@ -37,208 +40,236 @@
...
+    public Server(String clusterName, String mainDirectory, String languages, String transportAddresses,
+	    boolean isTest) {
+	try {
+	    if (SystemUtils.IS_OS_WINDOWS) {
+		setupDirectories(new URL("file:///" + mainDirectory));
+	    } else {
+		setupDirectories(new URL("file://" + mainDirectory));
+	    }
+	} catch (Exception e) {
+	    log.error("Can't create directories: ", e);
```
karussell, line 76: I know this is not your code ... but should we throw an exception instead of silently accepting the problem?

src/main/java/de/komoot/photon/query/ReverseRequestFactory.java
karussell: Also the check can be relative short via `expression? if-expression : else-expression`

src/main/java/de/komoot/photon/query/ReverseRequestFactory.java
```diff
@@ -14,26 +17,90 @@
...
+            }else{
+                // limit number of results to 50
+                limit = Math.min(limit, 50);
+            }
+        }
         
-        ReverseRequest reverseRequest = new ReverseRequest(location, language);
+                
+                
+        ReverseRequest reverseRequest = new ReverseRequest(location, language, radius, queryStringFilter, limit, locationDistanceSort);
```
karussell: (Here are also a few formatting issues)

src/test/java/de/komoot/photon/query/TagFilterQueryBuilderSearchTest.java
```diff
@@ -74,6 +73,8 @@ public void testFilterWithTagTourismAttraction() throws IOException {
         QueryBuilder queryBuilder = tagFilterQueryBuilder.buildQuery();
         SearchResponse searchResponse = search(client, queryBuilder);
         assertThat(searchResponse.getHits().getTotalHits(), is(2l));
+        
+        deleteIndex();
```
karussell, line 77: Maybe call deleteIndex in a new method `@After public void tearDown`?

src/test/resources/json_queries/test_base_query.json
```diff
@@ -1,102 +1,185 @@
...
+				"disable_coord": false,
+				"adjust_pure_negative": true,
+				"boost": 1.0	
+			}	
+		}],
+		"disable_coord": false,
+		"adjust_pure_negative": true,
+		"boost": 1.0	
+	}	
+}
```
karussell: formatting

src/test/resources/json_queries/test_base_query_fr.json
```diff
@@ -1,102 +1,185 @@
...
+				"disable_coord": false,
+				"adjust_pure_negative": true,
+				"boost": 1.0	
+			}	
+		}],
+		"disable_coord": false,
+		"adjust_pure_negative": true,
+		"boost": 1.0	
+	}	
+}
```
karussell: formatting

karussell:
There seems to be an issue with housenumbers:

The [query 'harburgerstraße 5 bayreuth'](http://46.4.67.134:2322/api?q=harburgerstra%C3%9Fe%205%20bayreuth) returns the correct one at position 3:

 * Harburgerstraße Bayreuth
 * Harburgerstraße 1 Bayreuth
 * Harburgerstraße 5 Bayreuth

(can send other problems if necessary)

hbruch:
Commit https://github.com/komoot/photon/pull/254/commits/6554c06e9e21924495298c30f1a2b7d55a8e582f performs much better for housenumbers.

However, concerning fuzzy matching, I encounter some unexpected results, e.g.:

Slighty varying [Alte Strasse, Ulm](http://46.4.67.134:2322/api?q=Alte%20Strasse,%20Ulm&lang=de) (with ss instead of ß) returns "Meininger Allee, Neu-Ulm" as first match vs. [expected photon.komoot.de result](http://photon.komoot.de/api?q=Alte%20Strasse,%20Ulm&lang=de)).

I think the formerly used asciifolding filter for search_ngram and index_ngram would improve matching in this case. 

For [Steigstr Stuttgart](http://46.4.67.134:2322/api?q=Steigstr%20Stuttgart) "Steinstraße, Stuttgart" is returned vs. [the expected result for photon.komoot.de](http://photon.komoot.de/api?q=Steigstr%20Stuttgart). 

@reuschling what is the rational behind the simplified index_ngram and search_ngram analyzers? It seems as there are a couple of edge cases for which the former filters improved matching.

karussell:
Ah, ok. Will trigger a new export.

BTW: @hbruch do you know a standard ES solution for e.g. Erlangerstraße vs. Erlanger Straße? (see issue #46) I find this currently the most annoying limitation in current photon

lonvia:
This PR now contains a rather large mixture of software updates, fixes and new features. This makes it rather difficult to see which changes to the database and is the result of which change to the code. I'd really like to see this PR split into multiple PRs with one for each issue, so it becomes easier to judge the impact of each change. Preferably the first PR would be one with the actual changes required for running on ES5 because it is a) likely uncritical and therefore can be quickly merged and b) everything else needs to build on it.

If the more than a thousand lines of remaining whitespace changes go away doing that, that would be a bonus.

karussell:
I'm not sure if there are new feature introduced (please correct me if I'm wrong). The problem is that the upgrade to ES 5.5 *requires* to change several old approaches and so breaking it into multiple smaller PRs is much harder. 

But to make sure we do not have regressions we need a good (and increasing) number of real world tests automated executed. E.g. https://github.com/geocoders/geocoder-tester and https://github.com/pelias/acceptance-tests but included in this repo. This issue from the pelias devs is an interesting read: https://github.com/pelias/acceptance-tests/issues/109 (see [this repo](https://github.com/pelias/fuzzy-tests) too)

lonvia:
Just reading the commit messages, there are at least three independent changes mixed in this PR:

* port to ES 5
* various changes to the reverse algorithm, including the new parameters limit and radius
* new request parameter 'distanceSort=true|false'

There is also the change to BM25 which may or may not be necessary for the ES5 port. I'd be intersted in the minimal set of changes required that gets us a database on ES5 as close as possible to the current one.

karussell:
> new request parameter 'distanceSort=true|false'

That is just an implementation detail and not a real new feature

BTW: we probably should not expose this parameter to the public, I think @reuschling mentioned that we can use it to maintain backward compatible but the new stuff should be better as it utilized new ES stuff

> various changes to the reverse algorithm, including the new parameters limit and radius

This is true (but also a very tiny part of the change)

> There is also the change to BM25

BM25 is now the default I think: https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules-similarity.html

karussell:
> Commit 6554c06 performs much better for housenumbers.

Indeed - now live!

karussell:
See #263 for a quality comparison and it seems there are no big regressions. One minor problem seems to be with French addresses, will comment there to print the differences or something. Also in #253 I'm currently only testing "address to coordinate" and no filtering or reverse coding stuff.

reuschling:
Most of the contents of the small json files were changed, but I can change the formatting to 'open bracket in the same line' anyway, due to the fact you don't like it.
The data size of the index has mostly changed because of five years ongoing Lucene development, especially new geo indices were introduced in this time. You have a magnitude of faster reverse geo lookups now - this is not for free, the price is enhanced and bigger data structures in the index.
@lonvia I made all the changes in this single commit for reason. This is not a small PR, but a whole migration to a complete new and redesigned version of the photon base technology. In this context, also the standard similarity of ES has changed from a cosine derivative to BM25. This resulted in a situation that the own-grown, hard coded weighted sum of the location bias didn't return reasonable results anymore. In the context to have a functional system after the migration, it was necessary to change this code, the solution was to sort against the distance to the query point, with an extra parameter to enable the old behavior, for backward compatibility.
@hbruch according to the ngram edge analyzer: I switched to the ngram edge tokenizer due to the reason that the wordendingautocomplete filter doesn't exist anymore in recent ES versions. According to Straße vs Strasse: this has nothing to do with the asscii folding, we could use

pom.xml
reuschling: I cannot fix this out of the box, I'm also doesn't sure if this is related to the migration. I have not the time to fix everything in photon in the context of the migration.

src/main/java/de/komoot/photon/App.java
reuschling: looks like tabs vs spaces

pom.xml
karussell: Sorry I do not meant it to sound this way. Before we did not see this warning and also log4j2 wasn't used. Now a proper initialization needs to be found (I can have a look into this).

src/main/java/de/komoot/photon/query/PhotonQueryBuilder.java
reuschling: it is a common prefix for a member variable

pom.xml
karussell: Just to explain my intention: I often use refactorings as a chance to clean up old code and instead of just leaving some ugliness  I'm used to fix this right away, but highly likely new issues are a much better place to fix them.

src/main/java/de/komoot/photon/elasticsearch/Server.java
karussell: (should go into new issue too :) )

karussell:
To me this sounds like a really good change and actual improvement. If the Straße vs Strasse problem is fixed I would vote to merge it. Would you too, @lonvia :) ?

reuschling:
I made a new commit, all should be included beside the logger config and the test case deleteIndex invocation

lonvia:
lang-painless.jar has been added in two different versions (5.4.1 and 5.5.0). Is that necessary?

As for paying a price for improved reverse geocoding: this is my point exactly. I see the strength of ES in forward geocoding. And I'm worried that those who only look for a search-as-you-type solution now need to pay with bigger hardware for a feature they don't need. That's my personal opinion, which may or may not be valid but I really think we should discuss this separately from the issue of porting to ES5, which definitely is a long overdue step.

karussell:
> And I'm worried that those who only look for a search-as-you-type solution now need to pay with bigger hardware for a feature they don't need. 

disc is cheap these days ;)

> but I really think we should discuss this separately from the issue of porting to ES5, which definitely is a long overdue step.

:+1: yes, we should discuss this in a separate issue as the functionality was there before.

karussell:
The updated index is now live and the ss vs. ß Problem is fixed. The other minor problem regarding the slightly inaccurate ranking for 'Steigstr Stuttgart' (2nd instead of 1st) mentioned from @hbruch is still present.

karussell:
As only the plugin version 5.5.0 is mentioned in the config I think we can remove 5.4.1

src/main/java/de/komoot/photon/query/PhotonRequestFactory.java
```diff
@@ -42,13 +53,25 @@ public PhotonRequestFactory(Set<String> supportedLanguages) {
         } catch (Exception nfe) {
             //ignore
         }
+        Boolean locationDistanceSort = true;
```
hbruch, line 56: I'd suggest to switch to 'false' as default. While I like the distance sort (=stronger bias) when zoomed to greater zoom levels, the former results are closer to my expectations when zoomed out. E.g. for [Stuttgart with a location bias of (51;9)](http://46.4.67.134:2322/api?q=Stuttgart&lang=de&limit=5&lat=51&lon=9) I would expect to see Stuttgart in the results, rather than 'Stuttgarter Straße' in Frankfurt or Kassel. Perhaps providing a zoom level or weighting factor for the bias would be an option for a later PR.

karussell: Hmmh, probably we should argue then how it was before (I think it was kind of broken ie it should be indeed false) because if you argue on expectation I can give you a counter argument like when you search for "bahnhof" I would expect the train station within that city.

> Perhaps providing a zoom level or weighting factor for the bias would be an option for a later PR.

Related to #198 ... probably we could also provide a bounding box filter.

karussell:
> And I'm worried that those who only look for a search-as-you-type solution now need to pay with bigger hardware for a feature they don't need.

Please @reuschling correct me if I'm wrong but I think even without reverse geocoding we need the geo index to provide the location bias.

reuschling:
@karussell yes indeed, geocoding is a must for searching and sorting against geo coordinates, even to provide the photon standard functionalities. @lonvia I removed the unnecessary jar, thanks for finding this. @hbruch, @karussell I would also go for the bounding box or a radius, to keep the sorting criteria well defined. We should also discuss this in a different issue.

reuschling:
@hbruch provided a nice solution for the 'Steigstr Stuttgart' issue. I merged it, thanks for this again.

karussell:
Thanks! I've deployed the recent change to the test server and ran the benchmark #263  - now the counts for every subtests are identical to the current master. (Haven't looked into the details though)

karussell:
The response times are worse than before except for France: poland 80sec, USA 54s, Germany 60s, France1 31s, France2 350s

But the single runs are sometimes different in itself - so I'll have to make the measurement more robust as a first step

karussell:
sorry for the bubble, was a measurement problem: Germany 18s, USA 17s, poland 6sec, France1 9s, France2 100s

karussell:
@lonvia do you agree or disagree regarding mergeability of this PR?

karussell:
Thanks @reuschling, @lonvia, @hbruch and finally the initiator @masda70!

karussell:
btw: @reuschling if you connect your git commit email to github, the commits will be linked to your account and you'll also appear as contributor in the stats.

imadk:
Is the radius parameter public? how would an api call look like?

karussell:
Please ask questions on the mailing list

PR #271, opened 2017-12-20
updated spark to 2.7.1 with minimal changes

karussell:
All tests run green. I think this is a very important upgrade due to the jetty dependency.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -108,7 +108,7 @@
 		<dependency>
 			<groupId>com.sparkjava</groupId>
 			<artifactId>spark-core</artifactId>
-			<version>1.1.1</version>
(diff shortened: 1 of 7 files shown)


.gitignore
```diff
@@ -5,7 +5,7 @@ dependency-reduced-pom.xml
 .idea/
 *.iml
 *.~*
-
+*~
```
lonvia: General note: I usually recommend to put editor-specific ignore patterns into $HOME/.gitignore because then they work for all projects and you don't have to rely on the maintainer to accept your favourite pattern.

karussell: Thanks, didn't know this!

PR #272, opened 2017-12-22
Fix #266

### LINKED ISSUE #266: Fix logging problems
As discussed in #254 there is a warning at startup and we need to test if we now get the full stacktrace when there is an exception while export.

hbruch:
This PR fixes the log configuration warning on startup.
Former dependency to log4j 1.2.7 is replaced by dependency to log4j-2.8.2 (as used by eleasticsearch 5.5.0).
photon itself uses slf4j which now delegates to the log4j implementation via the slf4j-log4j bridge.

The patternLayout now prints date and time and async appenders are used. Exceptions during export are logged.

I tried to use log4j2's AsyncLoggers without success. Possibly some the slf4j-log4j integration does not yet work with AsyncLoggers.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -11,7 +11,7 @@
 		<dependency>
 			<groupId>org.elasticsearch</groupId>
 			<artifactId>elasticsearch</artifactId>
-			<version>5.5.0</version>
+			<version>${elasticsearch.version}</version>
 			<exclusions><!-- gives a bad error -->
 				<exclusion>
 					<groupId>org.elasticsearch</groupId>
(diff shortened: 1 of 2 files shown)


karussell:
I still get for nominatim-photon export the following warning:

```
ERROR StatusLogger No log4j2 configuration file found. Using default configuration: logging only errors to the console. Set system property 'org.apache.logging.log4j.simplelog.StatusLogger.level' to TRACE to show Log4j2 internal initialization logging.
```

Did I do something wrong?

hbruch:
Hm. Not sure... How do you start the export? I tested the logging with the json-dump option enabled. Perhaps this is not the nominatim-photon export you had in mind?

karussell:
I'm using

```
java -jar /install/photon/target/photon-0.2.9-SNAPSHOT.jar -nominatim-import -host localhost -port xxx -database xxx -user xxx -password xxx -languages en,de,fr,it -data-dir $EXPORT_DIR
```

karussell:
BTW: The same problem occurs to me, for just starting up the search server too (with the latest or an old data snapshot)

hbruch:
Ah, I see... in the pom the resources path is explicitly set to "es", so src/main/resources is not copied to the jar. For me it seemed to work as eclipse copied the log4j2.xml in target/classes and I didn't do a clean build :-o

karussell:
Thanks a lot for finding this! Do you think we should avoid this and e.g. move the "es" stuff into src/main/resources or will this introduce other problems?

karussell:
Works as invented :) !

karussell:
Now I get "version [5.5.0-SNAPSHOT] is a pre-release version of Elasticsearch and is not suitable for production" although we have the dependency 5.5.0 defined. 

When I try to use 5.5.3 I get

IllegalArgumentException: plugin [lang-painless] is incompatible with version [5.5.3]; was designed for version [5.5.0]

and for 5.6.5 I get some undefined util class. Will investigate tomorrow.

hbruch:
It seems that elasticsearch >= 5.0 is [not intended to be run as embedded server](https://www.elastic.co/blog/elasticsearch-the-server). Elasticsearch detects it's [packaged in a shaded jar](https://discuss.elastic.co/t/using-embedded-version-of-elastic-gives-this-warning/84299/8), which results in the misleading SNAPSHOT warning, see also [Build.java](https://github.com/elastic/elasticsearch/blob/5.5/core/src/main/java/org/elasticsearch/Build.java#L72)

karussell:
Thanks for finding this! We discussed this with @reuschling some time ago and the only way it seems is to deprecate the embedded server and to make this external and install ElasticSearch before using photon. BTW: installing ES separately is not that hard, indeed we at GraphHopper prefer this since the beginning but it certainly adds a layer for new users.

PR #273, opened 2017-12-22
enable travis ci

karussell:


karussell:
Due to #274 we have to ignore failures for jdk9 for now

PR #279, opened 2018-01-05
recursive formatting incl. organize imports

### LINKED ISSUE #278: Use common white spaces across the project
I'm against tabs (for no reason ;) !), so I would like to use 4 spaces and apply also the same code formatting to all source files. Currently we have a wild mix it seems.

Using the standard java rules stored in IntelliJ, see [this discussion](https://github.com/graphhopper/graphhopper/issues/770#issuecomment-246490515)

karussell:
Fixes #278

### DIFF
--- a/src/main/java/de/komoot/photon/App.java
+++ b/src/main/java/de/komoot/photon/App.java
@@ -1,7 +1,6 @@
 package de.komoot.photon;
 
 
-
 import com.beust.jcommander.JCommander;
 import com.beust.jcommander.ParameterException;
 import de.komoot.photon.elasticsearch.Server;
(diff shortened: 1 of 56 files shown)


karussell:
Looks like auto import cleanup did too much. Will fix the tests.

lonvia:
That's quite large. But looks good so far. What do you think about adding a CONTRIBUTING.md with the styling rules to make it easier for contributors?

karussell:
Yes, good idea.

Regarding the large changes: I did just one click in the IDE and the changes are sometimes not properly detected in the github diff so this is a bit confusing at some places.

karussell:
> For this project, one single change like this one is better imho.

I think so too, because if just for touched files then there the changes are kind of invisible and so more error prone. This global reformatting (only to *.java btw) is automatically done and hopefully without a mistake :)

karussell:
Have created this new contributing guide here https://github.com/komoot/photon/blob/master/.github/CONTRIBUTING.md - please feel free to edit&discuss it

PR #286, opened 2018-01-09
added country code filter to export

### LINKED ISSUE #281: country extract
I would like to provide and build a smaller extract to use photon for testing purposes too. To achieve this we can filter the export by country, per @lonvia this should be as simple as adding
`WHERE country_code in ('de', 'ch', 'at')` in the two SQL queries in readEntireDatabase.
  
https://github.com/komoot/photon/commit/12b3f3f913203166785e581dba8f2aba78e7cdb9

karussell:
Fixes #281 

Will done an export for Switzerland, if ready, get it via 

```
wget -O - http://download1.graphhopper.com/public/photon-ch-180109.tar.bz2 | bzip2 -cd | tar x
```

I'll also do an export for France and Germany so that we can easier test it using the geocoder testing framework.

### DIFF
--- a/src/main/java/de/komoot/photon/nominatim/NominatimConnector.java
+++ b/src/main/java/de/komoot/photon/nominatim/NominatimConnector.java
@@ -305,7 +305,7 @@ public class NominatimConnector {
      * parses every relevant row in placex, creates a corresponding document and calls the {@link #importer} for every document
      */
     public void readEntireDatabase(String... countryCodes) {
-        log.info("start importing documents from nominatim ...");
+        log.info("start importing documents from nominatim (" + (countryCodes.length == 0 ? "global" : String.join(",", countryCodes)) + ")");
         final AtomicLong counter = new AtomicLong();
 
         final int progressInterval = 50000;
@@ -360,7 +360,7 @@ public class NominatimConnector {
                 }
             }
         });
-        
+
         template.query("SELECT place_id, osm_id, parent_place_id, startnumber, endnumber, interpolationtype, postcode, country_code, linegeo " +
                 " FROM location_property_osmline " +
                 whereCountryCodeStr +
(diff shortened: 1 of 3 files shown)


karussell:
Hmmh, it does not look like the filter is active as the process is now doing its work for over 1 hour.

src/main/java/de/komoot/photon/nominatim/NominatimConnector.java
```diff
@@ -353,7 +357,8 @@ public void processRow(ResultSet rs) throws SQLException {
 		});
 
 
-		template.query("SELECT place_id, osm_id, parent_place_id, startnumber, endnumber, interpolationtype, postcode, country_code, linegeo FROM location_property_osmline ORDER BY geometry_sector; ", new RowCallbackHandler() {
+		template.query("SELECT place_id, osm_id, parent_place_id, startnumber, endnumber, interpolationtype, postcode, country_code, linegeo FROM location_property_osmline " +
+				countryCodeStr + " ORDER BY geometry_sector; ", new RowCallbackHandler() {
```
lonvia: This looks wrong. It should be 'WHERE country_code ...' not 'AND country_code ...'.

src/main/java/de/komoot/photon/nominatim/NominatimConnector.java
```diff
@@ -321,8 +321,12 @@ public void readEntireDatabase() {
 		final BlockingQueue<PhotonDoc> documents = new LinkedBlockingDeque<PhotonDoc>(20);
 		Thread importThread = new Thread(new ImportThread(documents));
 		importThread.start();
+		String countryCodeStr = "";
+		if(countryCodes.length > 0)
+			countryCodeStr = " AND country_code in ('" + String.join("','", countryCodes) +"')";
```
lonvia: I would lower-case the strings as Nominatim expects them this way. I would also suggest to do a bit of sanity checking of the parts: length must be exactly two characters and characters must be ascii letters only.

karussell:
Thanks - will try your suggestion!

karussell:
Looks like this now worked (~3h for Germany)
```
wget -O - http://download1.graphhopper.com/public/photon-de-180109.tar.bz2 | bzip2 -cd | tar x
```

karussell:
Have tested this extract locally with success!

PR #289, opened 2018-01-11
avoid falling through catch and improve logs a bit

### LINKED ISSUE #267: Improve error handling
An exception should in the most cases not being catched and logged, instead rethrown as RuntimeException. There is this problem in Server, see discussion in #254 

In ReverseRequestFactory.java we do this too and catch a NPE which is ugly.

karussell:
Fixes #267

A few "falling through" catches were not fixed, e.g. when adding a document failed. I'm not sure if this is good, but I thought it might be that a single document fails and aborting the import process that takes hours seems ugly. But this way we silently accept potential bugs.

### DIFF
--- a/src/main/java/de/komoot/photon/App.java
+++ b/src/main/java/de/komoot/photon/App.java
@@ -121,11 +121,7 @@ public class App {
         de.komoot.photon.elasticsearch.Importer importer = new de.komoot.photon.elasticsearch.Importer(esNodeClient, args.getLanguages());
         NominatimConnector nominatimConnector = new NominatimConnector(args.getHost(), args.getPort(), args.getDatabase(), args.getUser(), args.getPassword());
         nominatimConnector.setImporter(importer);
-        try {
-            nominatimConnector.readEntireDatabase();
-        } catch (Exception e) {
-            throw new RuntimeException("error importing from nominatim: " + e.getMessage());
-        }
+        nominatimConnector.readEntireDatabase();
 
         log.info("imported data from nominatim to photon with languages: " + args.getLanguages());
     }
--- a/src/main/java/de/komoot/photon/App.java
+++ b/src/main/java/de/komoot/photon/App.java
@@ -77,8 +77,7 @@ public class App {
         try {
             esServer.recreateIndex();
         } catch (IOException e) {
-            log.error("cannot setup index, elastic search config files not readable", e);
-            return;
+            throw new RuntimeException("cannot setup index, elastic search config files not readable", e);
         }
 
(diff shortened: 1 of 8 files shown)


src/test/java/de/komoot/photon/query/ReverseRequestFactoryTest.java
```diff
@@ -19,8 +19,8 @@
     private ReverseRequest reverseRequest;
 
     public void requestWithLongitudeLatitude(Request mockRequest, Double longitude, Double latitude) {
-        Mockito.when(mockRequest.queryParams("lon")).thenReturn(longitude.toString());
-        Mockito.when(mockRequest.queryParams("lat")).thenReturn(latitude.toString());
```
karussell, line 23: This is a bit ugly that the tests need to know those implementation details

src/main/java/de/komoot/photon/elasticsearch/Server.java
```diff
@@ -60,25 +57,22 @@ public MyNode(Settings preparedSettings, Collection<Class<? extends Plugin>> cla
     }
 
     public Server(CommandLineArgs args) throws Exception {
-        this(args.getCluster(), args.getDataDirectory(), args.getLanguages(), args.getTransportAddresses(), false);
+        this(args.getCluster(), args.getDataDirectory(), args.getLanguages(), args.getTransportAddresses());
     }
 
-    public Server(String clusterName, String mainDirectory, String languages, String transportAddresses,
-                  boolean isTest) throws Exception {
+    public Server(String clusterName, String mainDirectory, String languages, String transportAddresses) throws Exception {
```
lonvia: Delete 'throws Exception'? Also above.

src/main/java/de/komoot/photon/App.java
```diff
@@ -126,7 +124,7 @@ private static void startNominatimImport(CommandLineArgs args, Server esServer,
         try {
             nominatimConnector.readEntireDatabase();
         } catch (Exception e) {
-            log.info("error importing from nominatim: " + e.getMessage());
+            throw new RuntimeException("error importing from nominatim: " + e.getMessage());
```
lonvia: You can remove the entire try-catch block. readEntireDatabase() only throws RuntimeExceptions in the first place.

karussell:
Have addressed the comments

karussell:
Would you mind to have another look @lonvia :) ?

karussell:
> Oops, I thought you had merged it already.

:) ... just wanted to make sure that we have at least 4 eyes to approve it :)

Issue #170, opened 2015-06-19
Nominatim import is hanging 


maraev:
Hi!
When i star import from nominatim db i get exeption like:

[main] ERROR de.komoot.photon.elasticsearch.Server - cannot install plugin: mobz
/elasticsearch-head: java.io.IOException: plugin directory c:\gis\photon\photon_
data\elasticsearch\plugins\head already exists. To update the plugin, uninstall
it first using --remove mobz/elasticsearch-head command
[main] ERROR de.komoot.photon.elasticsearch.Server - cannot install plugin: poly
fractal/elasticsearch-inquisitor: java.io.IOException: plugin directory c:\gis\p
hoton\photon_data\elasticsearch\plugins\inquisitor already exists. To update the
 plugin, uninstall it first using --remove polyfractal/elasticsearch-inquisitor
command
[main] INFO org.elasticsearch.node - [Ringmaster] version[1.4.1], pid[4056], bui
ld[89d3241/2014-11-26T15:49:29Z]
[main] INFO org.elasticsearch.node - [Ringmaster] initializing ...
[main] INFO org.elasticsearch.plugins - [Ringmaster] loaded [lang-mvel, wordendi
ng], sites [head, inquisitor]
[main] INFO org.elasticsearch.script - [Ringmaster] compiling script file [c:\gi
s\photon\photon_data\elasticsearch\config\scripts\general-score.mvel]
[main] INFO org.elasticsearch.script - [Ringmaster] compiling script file [c:\gi
s\photon\photon_data\elasticsearch\config\scripts\location-biased-score.mvel]
[main] INFO org.elasticsearch.node - [Ringmaster] initialized
[main] INFO org.elasticsearch.node - [Ringmaster] starting ...
[main] INFO org.elasticsearch.transport - [Ringmaster] bound_address {inet[/127.
0.0.1:9301]}, publish_address {inet[/127.0.0.1:9301]}
[main] INFO org.elasticsearch.discovery - [Ringmaster] photon/QbPSEUuYQiyuDWm0aH
R0tg
[elasticsearch[Ringmaster][clusterService#updateTask][T#1]] INFO org.elasticsear
ch.cluster.service - [Ringmaster] new_master [Ringmaster][QbPSEUuYQiyuDWm0aHR0tg
][WS-1882][inet[/127.0.0.1:9301]], reason: zen-disco-join (elected_as_master)
[main] INFO org.elasticsearch.http - [Ringmaster] bound_address {inet[/127.0.0.1
[...]

PR #290, opened 2018-01-11
Add hint about Nominatim DB user password

RhinoDevel:
See https://github.com/komoot/photon/issues/170#issuecomment-356892859.


karussell:
Thanks!

Issue #291, opened 2018-01-12
Provide json dump importer 


hbruch:
Photon already supports the export of json dump files in elasticsearch's import format via the '-json' flag.

As discussed in #288, photon should provide an option to import such an json dump file.

This would have a couple of advantages, like  
* speed up import times compared to the current nominatim import 
* allow us to provide region dumps
* easier setup of photon, i.e. for new contributors
* the possibility to supply custom data as json file

PR #293, opened 2018-01-12
(WIP) Add option to import json dump file (#291)

hbruch:
This is a first attempt to provide a json dump import (see #291). Importing a json dump on my machine was about 10x faster than importing from nominatim. 

However, there are some points I'd like to receive feedback on before doing a PR for merge:
* currently, no sanity checking is done on import, neither of the action line, nor the source line, should it?
* the JsonDumper does not export the PhotonDoc.uid. It should either be exported or the PhotonDoc should be restored from json so it can be recalculated from source
* JsonDumpConnector does not convert a source line into a PhotonDoc but bypasses the public Importer interface to add the source string. This is ugly, but avoids (unnecessary?) parsing and formatting of the source (but would require uid handling, see above)
* the elasticsearcg.Importer uses BulkRequestBuilder. Using [BulkProcessor](https://www.elastic.co/guide/en/elasticsearch/client/java-api/current/java-docs-bulk-processor.html) would provide concurrent updates, as well as retries with exponential backup in case of failures  (this should be another PR, I think)

### DIFF
--- a/src/main/java/de/komoot/photon/json/JsonDumpConnector.java
+++ b/src/main/java/de/komoot/photon/json/JsonDumpConnector.java
@@ -1,0 +1,66 @@
+package de.komoot.photon.json;
+
+import java.io.BufferedReader;
+import java.io.FileNotFoundException;
+import java.io.FileReader;
+import java.io.IOException;
+
+import de.komoot.photon.Connector;
+import de.komoot.photon.ImportProgressMonitor;
+import de.komoot.photon.elasticsearch.Importer;
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * Import json dump data.
+ *
+ * @author holger
+ */
+@Slf4j
+public class JsonDumpConnector implements Connector {
+
+    private BufferedReader reader;
+    private Importer importer;
+
+    public JsonDumpConnector(String filename) {
+        try {
+            this.reader = new BufferedReader(new FileReader(filename));
+        } catch (FileNotFoundException e) {
+            log.error("Json dump file '%1' not found", filename);
+        }
+    }
+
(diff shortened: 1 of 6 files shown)


src/main/java/de/komoot/photon/json/JsonDumpConnector.java
```diff
@@ -0,0 +1,66 @@
...
+public class JsonDumpConnector implements Connector {
+
+    private BufferedReader reader;
+    private Importer importer;
+
+    public JsonDumpConnector(String filename) {
+        try {
+            this.reader = new BufferedReader(new FileReader(filename));
+        } catch (FileNotFoundException e) {
+            log.error("Json dump file '%1' not found", filename);
```
karussell, line 28: Here we should throw a RuntimeException to avoid usage of this connector.

src/main/java/de/komoot/photon/json/JsonDumpConnector.java
```diff
@@ -0,0 +1,66 @@
...
+        }
+    }
+
+    public void setImporter(de.komoot.photon.Importer importer) {
```
karussell, line 32: shouldn't it be provided on construction and then importer and reader could be final?

karussell:
> currently, no sanity checking is done on import, neither of the action line, nor the source line, should it?

Sanity of which form like empty content?

> It should either be exported or the PhotonDoc should be restored from json so it can be recalculated from source

Which option do you prefer?

> This is ugly, but avoids (unnecessary?) parsing and formatting of the source (but would require uid handling, see above)

Can't judge about it. Both methods seems ok to me.

> this should be another PR, I think

Yes, speed is not important for the first implementation IMO

karussell:
Have included some improvements of this in #438

Issue #300, opened 2018-01-21
limit does not seem to work properly for location bias


karussell:
There seems to be something strange with the limit. If I use [limit=10](http://46.4.67.134:2322/api?q=Erlanger&lat=52.481944&lon=13.431129&limit=10) I sometimes get 7 hits, but using [limit=7](http://46.4.67.134:2322/api?q=Erlanger&lat=52.481944&lon=13.431129&limit=7) I only get 4? This seems to occur only for the location bias

hbruch:
This could be caused by a post-query street deduplication, which filters multiple objects of type highway with same name and postal_code.

karussell:
Ah, cool, didn't know this. If this is the case maybe we then use a heuristic like `*1.5` or `*2` to improve the situation a bit? And do you know where this happens?

hbruch:
Such a heuristic (applied for limits >1) could improve this a bit, yes.
The StreetDupesRemover is only used in [forward geocoding](https://github.com/komoot/photon/blob/master/src/main/java/de/komoot/photon/searcher/AbstractPhotonRequestHandler.java#L32).

karussell:
See #301

PR #301, opened 2018-01-22
increase limit to improve de-duplication count

karussell:
Improves #300


karussell:
Indeed, have fixed this.

PR #303, opened 2018-01-24
Fix #302: Read address attributes from nominatim

### LINKED ISSUE #302: Addresses in Mannheim's city centre have neither street nor place attribute
Photon returns seemingly correct matches for addresses in the city centre of Mannheim, Germany, which are addressed via the name of their place ("Quadrat"/square) instead of street.

E.g. http://photon.komoot.de/api?q=Mannheim,%20D3%204 returns the correct address, but has neither street nor place attribute.

In OSM, most of these osm buildings have a addr:street and addr:place attribute, as e.g. for [Mannheim, D3 4](https://www.openstreetmap.org/way/181662687).

As the address is found, I assume nominatim provides this information and photon should explicitly add it as source field.

Relates to #147

hbruch:
With this PR, the address key/values are read from nominatim and if it contains a street attribute, this is copied to via `street.put("name", addressStreet)`.

It fixes tests added with geocoders/geocoder-tester#38

### DIFF
--- a/src/main/java/de/komoot/photon/PhotonDoc.java
+++ b/src/main/java/de/komoot/photon/PhotonDoc.java
@@ -5,6 +5,7 @@ import com.vividsolutions.jts.geom.Envelope;
 import com.vividsolutions.jts.geom.Point;
 import lombok.Getter;
 import lombok.Setter;
+import lombok.extern.slf4j.Slf4j;
 
 import java.util.HashMap;
 import java.util.HashSet;
@@ -18,6 +19,7 @@ import java.util.Set;
  */
 @Getter
 @Setter
+@Slf4j
 public class PhotonDoc {
     final private long placeId;
     final private String osmType;
@@ -36,6 +38,8 @@ public class PhotonDoc {
     final private int rankSearch;
 
     private Map<String, String> street;
+    private Map<String, String> neighbourhood;
+    private Map<String, String> suburb;
     private Map<String, String> city;
     private Set<Map<String, String>> context = new HashSet<Map<String, String>>();
     private Map<String, String> country;
@@ -88,6 +92,8 @@ public class PhotonDoc {
         this.linkedPlaceId = other.linkedPlaceId;
         this.rankSearch = other.rankSearch;
         this.street = other.street;
+        this.neighbourhood = other.neighbourhood;
+        this.suburb = other.suburb;
         this.city = other.city;
         this.context = other.context;
         this.country = other.country;
@@ -127,10 +133,53 @@ public class PhotonDoc {
     public void completeFromAddress() {
         String addressStreet = address != null ? address.get("street") : null;
         if (addressStreet != null) {
-            if (street == null) {
-                street = new HashMap<>();
+            if (this.street == null) {
+                this.street = new HashMap<>();
             }
-            street.put("name", addressStreet);
+            setOrReplace(addressStreet, this.street, "street");
+        }
+        
+        String addressCity = address != null ? address.get("city") : null;
+        if (addressCity != null) {
+            if (this.city == null) {
+                this.city = new HashMap<>();
+            }
+            setOrReplace(addressCity, this.city, "city");
+        }
+        
+        String addressSuburb = address != null ? address.get("suburb") : null;
+        if (addressSuburb != null) {
+            if (this.suburb == null) {
+                this.suburb = new HashMap<>();
+            }
+            setOrReplace(addressSuburb, this.suburb, "suburb");
+        }
+        
+        String addressNeighbourhood = address != null ? address.get("neighbourhood") : null;
+        if (addressNeighbourhood != null) {
+            if (this.neighbourhood == null) {
+                this.neighbourhood = new HashMap<>();
+            }
+            setOrReplace(addressNeighbourhood, this.neighbourhood, "neighbourhood");
+        }
+        
+        String addressPostCode = address != null ? address.get("postcode") : null;
+        if (addressPostCode != null && !addressPostCode.equals(this.postcode)) {
+            if (log.isDebugEnabled()) {
+                log.debug("Replacing postcode "+this.postcode+" with "+ addressPostCode+ " for osmId #" + osmId);
+            }
+            this.postcode = addressPostCode;
+        }
+    }
+
+    private void setOrReplace(String name, Map<String, String> namesMap, String field) {
+        String existingName = namesMap.get("name");
+        if (!name.equals(existingName)) {
+            if (log.isDebugEnabled()) {
+                log.debug("Replacing "+ field +" name '"+existingName+"' with '"+ name+ "' for osmId #" + osmId);
+            }
+            // TODO: do we need to add former name to context or better not, as it might have been wrong?
+            namesMap.put("name", name);
         }
     }
 }
--- a/src/main/java/de/komoot/photon/PhotonDoc.java
+++ b/src/main/java/de/komoot/photon/PhotonDoc.java
@@ -38,8 +38,8 @@ public class PhotonDoc {
     final private int rankSearch;
 
     private Map<String, String> street;
-    private Map<String, String> neighbourhood;
-    private Map<String, String> suburb;
+    private Map<String, String> locality;
+    private Map<String, String> district;
     private Map<String, String> city;
     private Set<Map<String, String>> context = new HashSet<Map<String, String>>();
     private Map<String, String> country;
@@ -92,8 +92,8 @@ public class PhotonDoc {
         this.linkedPlaceId = other.linkedPlaceId;
         this.rankSearch = other.rankSearch;
         this.street = other.street;
-        this.neighbourhood = other.neighbourhood;
-        this.suburb = other.suburb;
+        this.locality = other.locality;
+        this.district = other.district;
         this.city = other.city;
         this.context = other.context;
         this.country = other.country;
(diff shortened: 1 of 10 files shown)


lonvia:
As I said in #302, this is not really a fix for Mannheim. Still, it's worth thinking about using the addr:* tag information now that Nominatim stores it properly. But if we go for that it should be a bit more complete.

A bit of background: there are two special addr tags: addr:street and addr:place. These are meant to create a connection between an OSM object and the house number in that address. If addr:street is there, then we have a standard address with street and housenumber. If addr:place is there then the house numbers are counted within some larger area. Normally, you would only use one or the other with an address (although there are, as always, exceptions. Look for 'conscription numbers').

Nominatim only uses these two addr: tags to find a 'base object' to each address and then computes the rest of the address parts by looking in which administrative areas the base object lies. The result is a kind of textual description of the location. Photon currently also uses only this information to create its response.

This location description is often the same as the postal address but not always. Here is where the other addr:* tags come into play. They normally should contain the real postal address of a place. So the content might be slightly different as to what Nominatim computes.

There are pros and cons to displaying postal address or location descriptions in the result. For photon you might actually end up with slightly better results using the addr:* tags because it does not show the entire hierarchy of admin boundaries and sometimes grabs the wrong one. The main disadvantage of the addr: tags is that in contrast to Nominatim's address information, there is no translation of place names.

To make a long story short: I think, if you start looking into addr:street, you should at least also look at addr:city, maybe addr:suburb, too. To get translations we might do something clever and use the content of the addr: tag to find the best match in Nominatim's address hierarchy. But I'm getting carried away.

hbruch:
Thank you, @lonvia for your very valuable insights. 

>  I think, if you start looking into addr:street, you should at least also look at addr:city, maybe addr:suburb, too. 

I added suburb and neighbourhood as fields. 

Currently, I'm just overwriting street, city, suburb, neighbourhood, postcode with the addr: attributes.
Some results I gathered from a test run around Mannheim with ~500.000 addresses:

Type | Count
-- | --
city name replacements (total) | 11681
city replacements, formerly not set | 22
city replacements, with Levenshtein <= 2 | 111
city replacements, with Levenshtein > 2 | 11548
street name replacements  (total) | 3977
street replacements, formerly not set | 2106
street replacements, with Levenshtein <= 2 | 853
street replacements, with Levenshtein > 2 | 1018

Replacements with Levenshtein <= 2 usually are caused by typos (e.g Schloßplatz vs. Schlossplatz vs. Schlosssplatz), hyphenation diffs (e.g. Fränkisch-Crumbach vs Fränkisch Crumbach), case diffs (Am Großen Wald vs. Am großen Wald).

Replacements with Levenshtein > 2 are caused by completly different names (e.g. Tairnbach -> Mühlhausen) or in a few cases added/missing location appendices/prepositions (e.g. Hirschhorn (Neckar) vs. Hirschhorn, Pariser Weg vs. Am Pariser Weg) or different abbreviations (e.g. Sankt-Leoner-Straße vs. St. Leoner Straße).

Postal codes seem not to be replaced, I assume Nominatim is already using addr:postcode in case it exists(?).

> To get translations we might do something clever and use the content of the addr: tag to find the best match in Nominatim's address hierarchy.

Currently, I do not yet use addr:* to pick the most probable address hierarchy match, and I'm not yet evaluating the addr:place tag. Instead of just replacing the default name with attr:* value, as I do now, another option would be to set them first and allow overwriting existing names with address hierarchy values, if  the address hierarchy default name matches exactly.

lonvia:
> Postal codes seem not to be replaced, I assume Nominatim is already using addr:postcode in case it exists(?).

Yes, you can leave that out.

As for the other tags: addr:street and addr:city are uncontested and addr:suburb is in moderately wide use as well. Beyond that it depends on the country, so I'd leave it at that for the moment and remove addr:neighbourhood.

src/main/java/de/komoot/photon/PhotonDoc.java
```diff
@@ -116,4 +126,60 @@ public boolean isUsefulForIndex() {
 
         return true;
     }
+    
+    /**
+     * Complete doc from nominatim address information.
+     */
+    public void completeFromAddress() {
+        String addressStreet = address != null ? address.get("street") : null;
```
lonvia: Check address == null once at the begin of the function and return early.

src/main/java/de/komoot/photon/PhotonDoc.java
```diff
@@ -116,4 +126,60 @@ public boolean isUsefulForIndex() {
...
+            setOrReplace(addressSuburb, this.suburb, "suburb");
+        }
+        
+        String addressNeighbourhood = address != null ? address.get("neighbourhood") : null;
+        if (addressNeighbourhood != null) {
+            if (this.neighbourhood == null) {
+                this.neighbourhood = new HashMap<>();
+            }
+            setOrReplace(addressNeighbourhood, this.neighbourhood, "neighbourhood");
+        }
```
lonvia, line 156: This code duplication can be avoided if you move the lookup into setOrReplace as well.

src/main/java/de/komoot/photon/nominatim/NominatimConnector.java
```diff
@@ -469,6 +471,16 @@ private void completePlace(PhotonDoc doc) {
...
 
+            if (address.isNeighbourhood() && doc.getNeighbourhood() == null) {
+                doc.setNeighbourhood(address.getName());
+                continue;
+            }
+
+            if (address.isSuburb() && doc.getSuburb() == null) {
+                doc.setSuburb(address.getName());
+                continue;
+            }
```
lonvia, line 500: I wouldn't recommend using the data from Nominatim's computed address at the moment because it is not reliable. Really only set suburb if it shows up in the address tags.

gopi-ar:
> addr:suburb is in moderately wide use as well. Beyond that it depends on the country, so I'd leave it at that for the moment and remove addr:neighbourhood.

At-least with Indian addresses, this doesn't seem to be the case.  A number of addresses on our street weren't returning suburb or neighbourhood with this PR.  `admin_level` seems reliable to some extent and all the linked place_ids in nominatim seemed to return correct admin_levels and were also marked as `boundary`. I added the following logic (based on admin_level info from [here](https://wiki.openstreetmap.org/w/index.php?title=Tag:boundary%3Dadministrative&uselang=en-US)  to @hbruch 's code and our office address in Hyderabad now shows the actual locality & suburbs.

````
    public boolean isNeighbourhood() {
        if ("place".equals(osmKey) && "neighbourhood".equals(osmValue)) {
            return true;
        }
        // TODO need better logic for this, not sure if it applies everywhere
        if (adminLevel != null && adminLevel == 11 && "boundary".equals(osmKey) && "administrative".equals(osmValue)) {
            return true;
        }
        return false;
    }

    public boolean isSuburb() {
        if ("place".equals(osmKey) && "suburb".equals(osmValue)) {
            return true;
        }

        // TODO need better logic for this, not sure if it applies everywhere
        if (adminLevel != null && (adminLevel == 9 || adminLevel == 10) && "boundary".equals(osmKey) && "administrative".equals(osmValue)) {
            return true;
        }
        return false;
    }
````

If this logic makes sense, will send an updated PR.

File: src/main/java/de/komoot/photon/nominatim/model/AddressRow.java [hunk 4]
```diff
@@ -32,6 +32,21 @@ public boolean isStreet() {
...
+        return false;
+    }
+    
+    public boolean isSuburb() {
+        if ("place".equals(osmKey) && "suburb".equals(osmValue)) {
+            return true;
+        }
+        // TODO admin?
+        return false;
+    }
```
lonvia, line 47: Similar to the other one, you should now use the rank address and get better results for that:

suburb: [17, 21]
neighbourhood: [22, 25]

src/main/java/de/komoot/photon/nominatim/NominatimConnector.java
lonvia: No longer an issue with the latest Nominatim.

src/main/java/de/komoot/photon/PhotonDoc.java
```diff
@@ -116,4 +126,62 @@ public boolean isUsefulForIndex() {
...
+    }
+
+    private void setOrReplace(String name, Map<String, String> namesMap, String field) {
+        String existingName = namesMap.get("name");
+        if (!name.equals(existingName)) {
+            if (log.isDebugEnabled()) {
+                log.debug("Replacing "+ field +" name '"+existingName+"' with '"+ name+ "' for osmId #" + osmId);
+            }
+            // TODO: do we need to add former name to context or better not, as it might have been wrong?
```
lonvia: I would do that. The address tag might have a typo and having the original improves the result.

leonardehrenfried:
@lonvia 

I incorporated you review feedback. I'd love it if you gave this another review. Thanks!

src/main/java/de/komoot/photon/nominatim/model/AddressRow.java
lonvia: Sorry, that was misleading. I meant inclusive ranges here. So it should be:

    return 22 <= rankAddress && rankAddress < 26;

and

    return 17 <= rankAddress && rankAddress < 22;

leonardehrenfried: Fixed that one.

simonpoole: Stylistically you should be returning the condition directly, that is
...
return "place".equals(osmKey) && "suburb".equals(osmValue);
...

Further I would suggest, as a rule, always adding javadoc to new methods.

leonardehrenfried: Hmm, GitHub doesn't update the view of the lines you have commented on. It actually already does what you suggest, @simonpoole.

And yes, I can add the javadoc.

src/main/java/de/komoot/photon/PhotonDoc.java
```diff
@@ -116,4 +126,62 @@ public boolean isUsefulForIndex() {
...
+        }
+    }
+
+    private void setOrReplace(String name, Map<String, String> namesMap, String field) {
+        String existingName = namesMap.get("name");
+        if (!name.equals(existingName)) {
+            if (log.isDebugEnabled()) {
+                log.debug("Replacing "+ field +" name '"+existingName+"' with '"+ name+ "' for osmId #" + osmId);
+            }
+            namesMap.put("formerName", existingName);
```
lonvia: I was more thinking of adding it to the context hashset. If you have a look at `completePlace()` function in `NominatimConnector.java`, you'll see that this is what happens when multiple city-like address parts show up.

src/main/java/de/komoot/photon/PhotonDoc.java
leonardehrenfried: I've removed _some_ duplication but not as much as I would have liked as the assignment to the instance fields (`this.locality`...) must happen outside `setOrReplace` or am I overlooking something?

leonardehrenfried:
@lonvia @simonpoole I've incorporated another round of review feedback. Please give it another look.

src/main/java/de/komoot/photon/PhotonDoc.java
lonvia: I was thinking along the lines of having a function which can be used like this:

    this.locality = extractAddress(this.locality, address, "neighbourhood")

Nevermind for now. It is good as is.

Issue #296, opened 2018-01-16
Increase offset for decay function


karussell:
As [discussed on the mailing list](https://lists.openstreetmap.org/pipermail/photon/2018-January/000097.html) we need a location bias that behaves a bit more like the old functionality and one possibility could be to increase the offset so that results "far away" are still considered with an enabled location bias.

karussell:
I think the problem is here `dist*1000`:

https://github.com/komoot/photon/pull/254/commits/6e89723e22c5bd9332bc03f8bb308b5781e19862

The new method `planeDistance` has meter as unit and so we need `dist/1000` instead!

karussell:
> Good catch! 

It still does not fix the location bias. I'm currently experimenting with a better script but this is far from easy to have a good balance and a zoomlevel dependent solution might be best :)

> Perhaps the newly introduced distance_sort=true then can be reverted, as suggested by @lonvia in #294?

The default behaviour is already fixed via #297.

karussell:
We really need better testing here. Is this possible with the geocoder-tester tool?

hbruch:
For [Ile'de France](https://github.com/geocoders/geocoder-tester/blob/master/geocoder_tester/world/france/iledefrance/test_location_bias.csv), there seem to be few test cases.

karussell:
Thanks, cool, that this is already possible! We have to add a few negative (where location bias has to be ignored) and positive (where location bias can be used for a better search)

lonvia:
The location bias should automatically be applied soon as the lat/lon parameters are added. I don't think we need an additional parameter there. A zoom parameter is a good idea as it is something users can understand (and automatically apply from the map that is shown). Nominatim uses a viewbox parameter and then has a binary bias of inside/outside that box. But that is not ideal either. Zoom sounds better.

What do you think about setting up an instance with an explicit location bias parameter that allows experimenting a bit?

karussell:
> What do you think about setting up an instance with an explicit location bias parameter that allows experimenting a bit?

I can do that, yes. Should I make the whole formula changeable via a URL parameter for the time being?

lonvia:
> I can do that, yes. Should I make the whole formula changeable via a URL parameter for the time being?

Yes, I think just setting the overall number for the location bias would be good. That allows to collect a few data points and then retrofit them to a formula.

karussell:
Now I am confused :) ... so complete access to a formula or just a distance as location bias?

lonvia:
[...]

PR #306, opened 2018-01-26
avoid distance_sort and use proper distance unit and tuned formula

karussell:
See discussion in #296


karussell:
Tests say that this is ok

except for

geocoder_tester/world/germany/test_location_bias.csv::Erlange

where now the expected result "Erlangen" comes 2nd and the 1st is "Berlin, Erlanger Straße" which is ok, due to the location bias being exactly there. Reducing the location_bias_scale to 1.6 would fix this, not sure what we want :)

karussell:
Now we have a problem for

`/api?q=bahnhof&lat=53.094488&lon=8.806057&limit=1&location_bias_scale=1.6`

and get a 'bahnhof' 40km away from the coordinate:

```
{"features":[{"geometry":{"coordinates":[9.1337314,53.0983492],"type":"Point"},"type":"Feature","properties":{"osm_id":240107774,"osm_type":"N","country":"Germany","osm_key":"place","city":"Ottersberg","osm_value":"suburb","name":"Ottersberg Bahnhof","state":"Lower Saxony"}}],"type":"FeatureCollection"}
```

src/main/java/de/komoot/photon/query/PhotonRequestFactory.java
```diff
@@ -19,7 +19,8 @@
     private final LanguageChecker languageChecker;
     private final static GeometryFactory geometryFactory = new GeometryFactory(new PrecisionModel(), 4326);
 
-    protected static HashSet<String> m_hsRequestQueryParams = new HashSet<>(Arrays.asList("lang", "q", "lon", "lat", "limit", "distance_sort", "osm_tag"));
+    protected static HashSet<String> m_hsRequestQueryParams = new HashSet<>(Arrays.asList("lang", "q", "lon", "lat",
+            "limit", "distance_sort", "osm_tag", "location_bias_scale"));
```
lonvia: Can distance_sort be removed here?

karussell: Ah, yeah had it removed locally already.

karussell:
> It would be particularly interesting to know what is a sensible range to choose here

Good question. This depends a bit on the search but probably anthing between 0.1 and 10

timautin:
Hello,

I'm trying to convert a zoom level to a location_bias_scale. Do you have an "ideal" solution? Is the new bbox parameter indirectly doing this?

Thanks!

PR #330, opened 2018-05-09
Remove diacritics from words

### LINKED ISSUE #326: Ignore diacritics in query
Currently Photon is quite fussy about diacritics ('accents'): 

* http://photon.komoot.de/api/?q=Ile+d%27Oleron returns only a track called "Voyeu de l'Ile d'Oleron"
* http://photon.komoot.de/api/?q=%C3%8Ele+d%27Ol%C3%A9ron returns the Île d’Oléron

Usually a placename search would ignore (or downrank) diacritic differences, so that a search for "Ile d'Oleron" still returned Île d’Oléron.

systemed:
Should fix #326. https://www.elastic.co/guide/en/elasticsearch/guide/current/asciifolding-token-filter.html explains the Elasticsearch behaviour.

### DIFF
--- a/es/index_settings.json
+++ b/es/index_settings.json
@@ -11,6 +11,7 @@
 					"preserving_word_delimiter",
 					"lowercase",
 					"german_normalization",
+					"asciifolding",
 					"unique"
 				]
 			},
@@ -21,7 +22,8 @@
 				"tokenizer": "standard",
 				"filter": [
 					"lowercase",
-					"german_normalization"
+					"german_normalization",
+					"asciifolding"
 				]
 			},
 			"index_raw": {


lonvia:
That took a couple of trials but here we go:

master: http://photon.staging.lonvia.de/master.html
this PR: http://photon.staging.lonvia.de/330.html

Both imported from the same Nominatim world data. At least the Ile d'Oleron example seems to do okay. Now I only need to find out how to run the geocoder tests.

systemed:
Superb - thanks for testing that, looks like it does what it should!

lonvia:
geocoding tester reports 26 new failures and 149 new passing. Full log at http://photon.staging.lonvia.de/330.log There are one or two odd ones like 'La Nouvelle Ève' where the exact match does not turn up anymore but altogether an improvement.

PR #331, opened 2018-05-17
Respect -cluster setting also when running standalone

lonvia:
Not sure if setting `-cluster` is the most proper way to go but it allows me to easily run multiple instances for comparison on the same machine.


karussell:
:+1:

PR #343, opened 2018-07-06
Added some details about nominatim updates on README

ThomasBarris:
Have updated the readme a bit since I asked myself what the update really does. 
It might be worth to think about the name of the shell script: continuously_update_from_nominatim.sh
It says it updates FROM nominatim but it also updates nominatim from OSM


lonvia:
Thanks! Even though you are right about the naming, I'm a bit reluctant to rename existing scripts. Your improvements to the README hopefully help already to make it a clearer what the script does.

PR #344, opened 2018-07-13
Tag query using filter. This improves caching for the same tags 

hendrikmoree:
There is only one filter (That can be used for caching) which included filtering on housenumber based on query. For every query, it's a different filter, so not useful for caching.

If you create a filter with only the list of tags, these may get cached and reused in every query. 
See https://www.elastic.co/guide/en/elasticsearch/guide/current/filter-caching.html

(No tests added, because nothing really changed. Only caching should be improved)


hbruch:
Hi @hendrikmoree , 
thanks for your PR! Nice spot! Looks good to me. Do you have any hints how to verify if ES now applies caching? And what improvements did you experience?

lonvia:
Updated in #423. Closing here.

PR #346, opened 2018-07-15
AdminLevel 10 is the real city name in the Netherlands

hendrikmoree:
If you search for http://photon.komoot.de/api/?q=dorpsstraat%20bennekom you get city name 'Ede' but that's incorrect.
The real city name is Bennekom, with this change you got the correct one in The Netherlands

Also see https://wiki.openstreetmap.org/wiki/Tag:boundary=administrative#10_admin_level_values_for_specific_countries for the admin_levels in The Netherlands

### DIFF
--- a/src/main/java/de/komoot/photon/nominatim/NominatimConnector.java
+++ b/src/main/java/de/komoot/photon/nominatim/NominatimConnector.java
@@ -446,12 +446,12 @@ public class NominatimConnector {
                 doc.setPostcode(address.getPostcode());
             }
 
-            if (address.isCity()) {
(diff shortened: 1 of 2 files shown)


lonvia:
I'm closing this here as we need a more generic solution to the problem. The Netherlands is not the only country where the address parts differ from the generic assignments and we don't want to have tons of hard-coded exceptions. I'm thinking of a generic solution in Nominatim at the moment, see [rank configuration](http://nominatim.org/release-docs/develop/develop/Ranking/#rank-configuration).

Issue #268, opened 2017-11-06
Provide distance or bounding box filter


karussell:
Related to #198 discussed in #254

otbutz:
Maybe a good candidate for the 0.4.0 milestone?

lonvia:
Fixed in #364.

PR #364, opened 2018-10-12
Add bbox parameter

hbruch:
Introduce a bbox to filter results contained in a given bounding box as suggested by #268.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,7 @@ Photon software is open source and licensed under [Apache License, Version 2.0](
 - location bias
 - typo tolerance
 - filter by osm tag and value
+- filter by bounding box
 - reverse geocode a coordinate to an address
 - OSM data import (built upon [Nominatim](https://github.com/twain47/Nominatim)) inclusive continuous updates
 
@@ -130,6 +131,11 @@ http://localhost:2322/api?q=berlin&limit=2
 #### Adjust Language
 ```
 http://localhost:2322/api?q=berlin&lang=it
+```
+
+#### Filter results by bounding box
+```
+http://localhost:2322/api?q=berlin&bbox=9.5,51.5,11.5,53.5
 ```
 
 #### Filter results by [tags and values](http://taginfo.openstreetmap.org/projects/nominatim#tags) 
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -132,6 +133,11 @@ http://localhost:2322/api?q=berlin&limit=2
 http://localhost:2322/api?q=berlin&lang=it
 ```
 
+#### Filter results by bounding box
+```
+http://localhost:2322/api?q=berlin&bbox=9.5,51.5,11.5,53.5
```
karussell, line 139: It would be important to mention which order the bbox has. Maybe we should also use a different order with left,bottom and right,top points ala `latSouth,lonWest,latNorth,lonEast` or why did you pick this order?

hbruch: You are right, this should be documented. Concerning the order: to me this was the one I‘m used to, but [Nomination and Overpass](https://wiki.openstreetmap.org/wiki/Bounding_Box) both use a different order.
Should I change it?
And before merging, I should add a verification of the lat/lon limits.

karussell: > Should I change it?

lon,lat,... is not a wrong choice (that's the order geojson uses I think) but the point parameter is lat,lon order and so it might be a bit inconsistent, but I'm not sure and probably not super important.

karussell: > Should I change it?

Up, not necessary. I confused the APIs ;)

lonvia: Did you change the order now? Because I read from the code below that you expect <minLon, maxLon, minLat, maxLat>. My preference would be as documented: <minLon,minLat,maxLon,maxLat>. (Or to be precise: <x1,y1,x2,y2>, Envelope() does not seem to care if it is min,max or max,min.)

Just to mention it, the odd boundingbox format of Nominatim is only used for output. The viewbox input parameter also expects <minLon,minLat,maxLon,maxLat>.

karussell:
> Stupid question: how does this work well together with the location bias?

Should work with it. E.g. it should be still possible to have a bias inside a bounding box.

README.md
karussell: As you have a preference then we should do x1,y1,x2,y2

lonvia:
Closing in favour of #421.

Issue #337, opened 2018-06-07
DocumentMissingException when running nominatim-update


privatwolke:
I am updating the photon index daily from a Nominatim instance. When calling `/nominatim-update` I regularly see messages like this:

```
2018-06-07 02:40:52,583 [Thread-13] ERROR de.komoot.photon.elasticsearch.Updater - error while bulk update: failure in bulk execution:
[34]: index [photon], type [place], id [1571637], message [[photon/mwDug3BeRs6nLwJoZhQHWA][[photon][2]] DocumentMissingException[[place][1571637]: document missing]]
```

The "document missing" line appears several hundred times for different documents. I am also unsure whether the update process finishes -- should there be a message in the console?

lonvia:
Fixed in #382.

PR #369, opened 2018-10-17
Fixed the update process by also complete places before updating

hendrikmoree:
Currently updates aren't correctly processed. Places are incomplete, and a lot warning are given because a place can be deleted and also tried to updateOrCreate afterwards.

This fixes both issues.

### DIFF
--- a/src/main/java/de/komoot/photon/nominatim/NominatimUpdater.java
+++ b/src/main/java/de/komoot/photon/nominatim/NominatimUpdater.java
@@ -39,8 +39,13 @@ public class NominatimUpdater {
                 for (UpdateRow place : getIndexSectorPlaces(rank, (Integer) sector.get("geometry_sector"))) {
 
                     template.update("update placex set indexed_status = 0 where place_id = ?", place.getPlaceId());
-                    final List<PhotonDoc> updatedDocs = exporter.getByPlaceId(place.getPlaceId());
 
+                    if (place.getIndexdStatus() == 100) {
+                        updater.delete(place.getPlaceId());
+                        continue;
+                    }
+
+                    final List<PhotonDoc> updatedDocs = exporter.getByPlaceId(place.getPlaceId());
                     for (PhotonDoc updatedDoc : updatedDocs) {
                         switch (place.getIndexdStatus()) {
                             case 1:
@@ -53,9 +58,6 @@ public class NominatimUpdater {
                                     break;
                                 }
                                 updater.updateOrCreate(updatedDoc);
-                                break;
-                            case 100:
-                                updater.delete(place.getPlaceId());
                                 break;
                             default:
                                 LOGGER.error(String.format("Unknown index status %d", place.getIndexdStatus()));
--- a/src/main/java/de/komoot/photon/nominatim/NominatimUpdater.java
+++ b/src/main/java/de/komoot/photon/nominatim/NominatimUpdater.java
@@ -39,25 +39,28 @@ public class NominatimUpdater {
                 for (UpdateRow place : getIndexSectorPlaces(rank, (Integer) sector.get("geometry_sector"))) {
 
                     template.update("update placex set indexed_status = 0 where place_id = ?", place.getPlaceId());
-                    final PhotonDoc updatedDoc = exporter.getByPlaceId(place.getPlaceId());
+                    final List<PhotonDoc> updatedDocs = exporter.getByPlaceId(place.getPlaceId());
 
-                    switch (place.getIndexdStatus()) {
-                        case 1:
-                            if (updatedDoc.isUsefulForIndex())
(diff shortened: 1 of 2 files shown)


hendrikmoree:
This fixes #337

src/main/java/de/komoot/photon/nominatim/NominatimUpdater.java
```diff
@@ -39,25 +39,28 @@ public void update() {
...
-                            updater.delete(place.getPlaceId());
-                            break;
-                        default:
-                            LOGGER.error(String.format("Unknown index status %d", place.getIndexdStatus()));
-                            break;
+                                break;
+                            default:
+                                LOGGER.error(String.format("Unknown index status %d", place.getIndexdStatus()));
+                                break;
+                        }
```
lonvia, line 65: Looking at this again, the different cases need a bit more consideration.

For 'case 100' (object was deleted), export.getByPlaceId() shouldn't even be called because we already know that the placeID won't exist anymore. (That was wrong before as well but is worth fixing while changing this code.)

More importantly, for the 'case 2', the code does not catch when a housenumber gets deleted from the batch of housenumbers for that placeID because the code only updates the ones that exist after the update. The most simple solution here is to always call a delete with a subsequent create for case 2. I don't know what the performance impact of such a solution would be.

In any case, the delete() in 'case 2' should be moved out of the loop. `isUsefulForIndex()` evaluates always to the same value for the elements of `updatesDocs` and calling delete() multiple times for the same placeID is inefficient, if not harmful.

hendrikmoree: I've fixed the case '100' issue. 

I'm not really sure how to fix the other case '2' issue. We're already running some months with this code, so it doesn't seems to be harmful. This PR looks quite important to me, because the update process is really broken now, if anyone now how to fix it, it would be great.

simonpoole: I think @lonvia wanted something along the lines of
```

                   final List<PhotonDoc> updatedDocs = exporter.getByPlaceId(place.getPlaceId());
                    if (place.getIndexdStatus() == 2 && updatedDocs.size() > 1) {
                        updater.delete(placeId);
                    }
                    for (PhotonDoc updatedDoc : updatedDocs) {
                        switch (place.getIndexdStatus()) {
                        case 1:
                            if (updatedDoc.isUsefulForIndex()) {
                                updater.create(updatedDoc);
                            }
                            break;
                        case 2:
                            if (updatedDoc.isUsefulForIndex()) {
                                updater.updateOrCreate(updatedDoc);
                            }
                            break;
```

While not elegant it should at least be correct, what the current code isn't.

simonpoole: @hendrikmoree I think @lonvia was thinking about something along

```
                    final List<PhotonDoc> updatedDocs = exporter.getByPlaceId(place.getPlaceId());
                    if (indexdStatus == 2 && updatedDocs.size() > 1) {
                        updater.delete(placeId);
                    }
                    boolean wasUseful = false;
                    for (PhotonDoc updatedDoc : updatedDocs) {
                        switch (indexdStatus) {
                        case 1:
                            if (updatedDoc.isUsefulForIndex()) {
                                updater.create(updatedDoc);
                            }
                            break;
                        case 2:
                            if (updatedDoc.isUsefulForIndex()) {
                                updater.updateOrCreate(updatedDoc);
                                wasUseful = true;
                            }
                            break;
                        default:
                            LOGGER.error(String.format("Unknown index status %d", indexdStatus));
                            break;
                        }
                    }
                    if (indexdStatus == 2 && !wasUseful) {
                        updater.delete(placeId);
                    }
```
This is better, but still has a couple of cases that are not covered, for example that the object now only has 1 house number and used to have multiple. 

To make it really correct you would need to remove the updatedDocs.size() > 1 test, which means that every update would first delete all documents with the referenced place id,

simonpoole:
@lonvia @karussell my question here would be: do we really need the added complexity caused by expanding Nominatim objects with multiple house numbers in to multiple documents on import in the first place?

lonvia:
They are needed to support address interpolations.

simonpoole:
There are two issues that are not addressed by my rough code suggestion

- interpolations were not being updated at all (this is simply an omission in the current code)
- you can't call ...updateOrCreate after calling delete as, updateOrCreate checks if the documents exist to determine if to call update or create when submitting the action for batch execution, and as a result update will incorrectly be called instead of create.  

I now have (I believe :-)) working code for this that I'll refine a bit more (are there any tests for the updater?) and then create a new PR if nobody disagrees.

lonvia:
Superseded by #382. Closing here.

Thanks @hendrikmoree for the initial version.

Issue #81, opened 2014-07-31
CORS Basic Support


TheRealPanPan:
Hi,

I'm using photon in a cross domain way.
I've just add 
response.header("Access-Control-Allow-Origin","*");
in app.java inside the api route handler.
This enable photon to answer to CORS request by accepting any origin to access its ressources (I.E. the response of ?q= queries).

A more in depth support would be to have these authorized domains to be configurable.

I Hope this helped.

Regards,
TheRealPanPan

Defozo:
Any update on this?

PR #373, opened 2018-11-20
CORS added for cross-site requests and readme updated

rahulhaque:
One extra args `-cors enable|disable` added to control cross-site request access. Updated readme for feature acknowledgment. Also updated related tests for extra args added.

This solves the issue [#81](https://github.com/komoot/photon/issues/81) and other CORS related errors as far as my test goes.


lonvia:
Now merged as #398. Many thanks, @rahulhaque for the initial implementation. Note that the parameter name has changed slightly in the merged version. It should be `-cors-any` now.

PR #377, opened 2018-12-28
Index ref, left, right and short 'names'

hendrikmoree:
Some name variants aren't indexed. This also adds these properties

### DIFF
--- a/es/mappings.json
+++ b/es/mappings.json
@@ -377,6 +377,34 @@
 						"copy_to": [
 							"collector.default"
 						]
+					},
+					"ref": {
+						"type": "text",
+						"index": false,
+						"copy_to": [
+							"collector.default"
+						]
+					},
+					"left": {
+						"type": "text",
+						"index": false,
+						"copy_to": [
+							"collector.default"
+						]
+					},
+					"right": {
+						"type": "text",
+						"index": false,
+						"copy_to": [
+							"collector.default"
+						]
+					},
+					"short": {
+						"type": "text",
+						"index": false,
+						"copy_to": [
+							"collector.default"
+						]
 					}
 				}
 			},
--- a/src/main/java/de/komoot/photon/Utils.java
+++ b/src/main/java/de/komoot/photon/Utils.java
@@ -93,6 +93,18 @@ public class Utils {
         if (name.get("reg_name") != null)
             fNames.put("reg", name.get("reg_name"));
 
+        if (name.get("short_name") != null)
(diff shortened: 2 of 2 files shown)


lonvia:
Sorry for the late reply.

So, the left and right variants should be fine, no objection there.

`ref` tends to be problematic in my experience as it interferes in strange ways with the other search terms. Might be different with the analyser we use but I'd like to see some tests for that.

I'm not sure that photon needs `name:short`. The main uses are that it contains a partial variant of `name` which the index should then have already or it contains a reference for which the objections above apply.

hendrikmoree:
I think `name:short` and `short_name` adds useful extra information for example on this nodes: https://www.openstreetmap.org/node/588503900
https://www.openstreetmap.org/way/547703999
This short variant of the same name wont be indexed by the current Photon because it's more of an abbreviation

I see i've included `short_name` in this PR, not `name:short`. But i think both will be usefull

Cliveptr:
I'd also like to see short_name used, though I'm not familiar with how it would interfere with other search terms. Without short_name you have to remember the full name of some locations, names that are often never used in reality.

b3nn0:
Is the ref-indexing still problematic? Because this seems to be quite a significant issue, or am I missing something?
E.g. https://photon.komoot.io/api/?q=K%208279&lang=de does certainly not find the "K 8279"
https://www.openstreetmap.org/search?query=48.057746%2C9.115506#map=16/48.0574/9.1150

The same issue exists for reverse.. It will simply return no usable road name at all:
https://photon.komoot.io/reverse?lat=48.057746&lon=9.115506

Is that something that is planned to be resolved? Feels like quite a severe limitation...

CutestNekoAqua:
yea short_names should be indexed.. example BER short_name for Berlin Brandenburg Airport. Many people use BER instead of the full name because.. I dunno tbh but photon obv just fails finding it for "BER" atm lol

laem:
Hi, just found this PR looking for why "short_name" isn't indexed. 

In France, the Direction du Numérique (French state's head of public IT) is well known by its acronym DINUM. 

Lots of places, especially administrative places are known by their acronym.

CutestNekoAqua:
> Hi, just found this PR looking for why "short_name" isn't indexed.
> 
> In France, the Direction du Numérique (French state's head of public IT) is well known by its acronym DINUM.
> 
> Lots of places, especially administrative places are known by their acronym.

same in germany :)

CutestNekoAqua:
@lonvia what are your thoughts on indexing these?

PR #382, opened 2019-01-24
Improvements to updating code

simonpoole:
- expand documents for multiple house numbers on objects and address
interpolations, based on code from @hendrikmoree
- retrieve and process updates to interpolations
- avoid inconsistencies in housenumber updates by deleting existing
documents
- remove code that tried to set indexed_status to 0 in the source
Nominatim database (running update.php --index will do that)
- generate some minimal stats
- minimal code cleanup

### DIFF
--- a/src/main/java/de/komoot/photon/nominatim/NominatimUpdater.java
+++ b/src/main/java/de/komoot/photon/nominatim/NominatimUpdater.java
@@ -12,6 +12,7 @@ import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.locks.ReentrantLock;
 
 /**
  * Nominatim update logic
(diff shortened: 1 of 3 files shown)


lonvia:
Could you get rid of the white-space-/line-break-only changes to make review a bit easier?

simonpoole:
Should be better now. For NominatimConnector the reformatting was accidental, as it was obvious that the slightly "unconventional" existing formatting would cause havoc, but "Macht der Gewohnheit".

simonpoole:
> 
> 
> Those "if" cascades make my head hurt. But looks okay so far. I need to find a minute to test this.
> 
I simplified the counting logic a tiny bit to make it more palatable.
> Am I right in assuming that you have not done the update of indexed_state yet or have I overlooked something?

The updating of the Nominatim database is on lines 49, 105 and 127. Change on line 179 so that the updates actually do something.

lonvia:
Planet test is running. It will take a week or so. I need to create an updatable Nominatim DB first.

lonvia:
Test instance finally running (and updating) at http://photon.staging.lonvia.de/382.html

The continuously_update_from_nominatim.sh still doesn't work properly. Nominatim's index script correctly detects that photon hasn't finished but then the script runs the photon update again anyway which leads to strange outcomes (multiple updates in photon running in parallel). Can photon detect by itself that another update is still running and refuse to start a second one or do we have to built in something into the script?

simonpoole:
Hmm yes, there currently isn't a lock of any kind that would stop that from happening, shouldn't be an issue though to fix.

lonvia:
Thanks, will try that.

Meanwhile another issue came up: if there are no documents to update I get:
```
2019-02-17 17:53:28,844 [Thread-10] INFO  de.komoot.photon.nominatim.NominatimUpdater - 0 interpolations created or updated, 0 deleted, 0 documents added or updated
Exception in thread "Thread-10" org.elasticsearch.action.ActionRequestValidationException: Validation Failed: 1: no requests added;
        at org.elasticsearch.action.ValidateActions.addValidationError(ValidateActions.java:29)
        at org.elasticsearch.action.bulk.BulkRequest.validate(BulkRequest.java:600)
        at org.elasticsearch.action.support.TransportAction.execute(TransportAction.java:131)
        at org.elasticsearch.action.support.TransportAction.execute(TransportAction.java:84)
        at org.elasticsearch.client.node.NodeClient.executeLocally(NodeClient.java:83)
        at org.elasticsearch.client.node.NodeClient.doExecute(NodeClient.java:72)
        at org.elasticsearch.client.support.AbstractClient.execute(AbstractClient.java:408)
        at org.elasticsearch.action.ActionRequestBuilder.execute(ActionRequestBuilder.java:80)
        at org.elasticsearch.action.ActionRequestBuilder.execute(ActionRequestBuilder.java:54)
        at de.komoot.photon.elasticsearch.Updater.updateDocuments(Updater.java:64)
        at de.komoot.photon.elasticsearch.Updater.finish(Updater.java:30)
        at de.komoot.photon.nominatim.NominatimUpdater.update(NominatimUpdater.java:126)
        at de.komoot.photon.App.lambda$null$0(App.java:152)
        at java.lang.Thread.run(Thread.java:748)
```

I guess it can be ignored but it's not pretty. I have also seen a NullPointerException at this point but I suspect that was still from the screwed up state after running multiple updates in parallel.

simonpoole:
@lonvia I fixed that in https://github.com/komoot/photon/pull/385/commits/b602fd4357caa1f190861cb62a377eeb9021e6e0 in the other PR.

src/main/java/de/komoot/photon/nominatim/NominatimUpdater.java
```diff
@@ -29,135 +30,148 @@
...
+                        deletedInterpolations, interpolationDocuments));
+                updater.finish();
+                template.update("update import_status set indexed=true;"); // indicate that we are finished
+
+                LOGGER.info("Finished updating");
+            } finally {
+                updateLock.unlock();
             }
+        } else {
+            LOGGER.info("Update already in process");
```
lonvia: Do you mean 'progress'?

simonpoole:
It was something around 3am:-). Will fix.

simonpoole:
I've fixed the typo and moved the commit that suppresses the error on empty updates to this PR.

lonvia:
The locking works like a charm. The database is now catching up at a rate of about 1 day's worth updates every 2 hours. I'll leave it running for a day or two to see if other issues come up.

Wrt empty updates. Does it still set the indexed column in import_status to true? That was the more serious problem above. Having no data to update and indexed = false, both nominatim and photon refused to do any work until I manually reset the table.

simonpoole:
... 
> Wrt empty updates. Does it still set the indexed column in import_status to true? That was the more serious problem above. Having no data to update and indexed = false, both nominatim and photon refused to do any work until I manually reset the table.

Yes, Updater.finish simply returns without doing anything, so the SQL should always get executed. One could argue that https://github.com/komoot/photon/pull/382/files#diff-e9d2005193ee859b90ddc8de7035194eR135 belongs in the finally clause so that it will always be executed even if something throws an Error or Exception, but on the other hand having everything come to a standstill if something goes wrong is probably preferable.

lonvia:
> but on the other hand having everything come to a standstill if something goes wrong is probably preferable.

Yes, definitely. It would be even better if would fail more loudly then just silently loop forever but I have no idea how to achieve that with the asynchronous update process. Something for a different PR in any case.

lonvia:
Updates have been running without issues for two days now and the database has managed to catch up on a full month of data. So we are good. Thanks a lot for looking into this.

PR #385, opened 2019-01-28
Add replication file generation and replication from file [WIP]

simonpoole:
This adds a facility to generate replication files in the "classical" OSM directory layout style and a facility to consume such files from a remote location via http.

New options:

Replication files generation:   
    **-replication-dir** dir _directory for replication files (default '.')_
    **-generate-files** _generate replication files while updating from Nominatim_
    **-replicate** _generate a single replication file without starting the server_

    
Consuming replication files:
    **-replication-url** url _url of base directory for file based replication (default none)_
    **-replication-interval** number _interval between trying to read replication files in minutes (default 60)_

This is work in progress and will need some testing irl.


lonvia:
I've played with hourly planet diffs. The resulting files are about half the size of the original diff, which is good.

Looking at the format, we need to get rid of any traces of Nominatim place IDs as they are tied to instances and it must be possible to create a continuous stream of diffs even when the producing instance changes. The ids should be fairly easy to replace. A string of "<osm type><osm id><class>" uniquely identifies a document. The other fields that contain place ids could preferably be just left out of the dump. They are only intermediate values that are needed to compute the address fields but they are not used in the ES document (although they seem to be imported).

lonvia:
I'm finally closing this as stale.

The new import/export function should make it easy to implement something similar now. The main open point is the ID handling, where we'd need support for place_id-less import files.

That said, offering replication files would be a fairly large task in terms of setting up the infrastructure. So this is going to be one of the topics I'm open to see implemented but that will only happen if there is support in form of funding or man-power.

PR #387, opened 2019-02-03
make -data-dir less confusing

kannes:
explicitly mention that -data-dir must point to *parent* of photon_data


kannes:
Because I just spent too much time being completely lost :o)

lonvia:
We actually might want to make that a little bit more consistent in general at some point. But for now, documenting it is a good start.

Issue #227, opened 2016-06-08
No result if country included


karussell:
[This query](https://photon.komoot.io/api/?q=%C3%96stra%20Ringw%C3%A4gen%20Tidaholm%2052226) results in the correct match but when I include the country 'Sweden', which is also in the response then I get no response. Is this a bug?

hbruch:
Interesting. If I [add Sweden in Swedish (Sverige)](http://photon.komoot.de/api/?q=%C3%96stra%20Ringw%C3%A4gen%20Tidaholm%2052226%20Sverige), the result is found.

Seems to be an issue with locale handling.

lonvia:
#538 introduces mixed-language queries for matching results. The original query has two "typos": Ringvägen" is written with "w" and the postcode is wrong. That brings us in the region of fuzzy search which is not yet available as mixed-language.

lonvia:
This works okay now in the OpenSearch version which is a bit more clever about mixed language fuzzy search.

PR #391, opened 2019-03-14
Fix for the house number cause the erratic behaviour

DavidGorritiGeograma:
e.g. we received a golf course hole in Australia instead of an address in Austria as the following:
http://photon.komoot.de/api/?q=3%20Universitaetsplatz,%20graz,%20austria

With location_bias_scale, it still shows the first two results in Italy:
http://photon.komoot.de/api/?q=3%20Universitaetsplatz,%20graz,%20austria&lat=47.58472242&lon=14.13731478&location_bias_scale=10

However, the result is correct without house number:
http://photon.komoot.de/api/?q=Universitaetsplatz,%20graz,%20austria

Therefore, we increase the prefix matching number, and boost the city and street name to increase the rank of the matching location names.
And modify the bias boost mode to locate the result closer to the bias center.

Thanks  for your review and help!
David Gorriti


src/main/java/de/komoot/photon/query/PhotonQueryBuilder.java
```diff
@@ -50,29 +50,32 @@
     private MatchQueryBuilder defaultMatchQueryBuilder;
 
     private MatchQueryBuilder languageMatchQueryBuilder;
+    
+    private MatchQueryBuilder numricMatchQueryBuilder;
```
simonpoole, line 60: This variable isn't used anywhere , and should be removed or used (PS it would be "numeric").

DavidGorritiGeograma: Thank you very much! I removed the unused the parameter. By the way, could you please help me to review if there is a better solution to solve the issue? Thanks again!
David

simonpoole: We are rather thin on ES knowledge here and up to now none of the relevant experts has weighed in yet, but, speaking for myself, the main question is how much this impacts the behaviour outside of the specific case you've been looking at (which it clearly improves). In my preliminary tests the change in behaviour is large so it will take some time to get to a definite conclusion on the PR.

DavidGorritiGeograma: Thank you so much for your explanation and help!

src/main/java/de/komoot/photon/query/PhotonQueryBuilder.java
```diff
@@ -84,12 +87,13 @@ private PhotonQueryBuilder(String query, String language) {
 
         m_finalQueryWithoutTagFilterBuilder = new FunctionScoreQueryBuilder(m_query4QueryBuilder, m_alFilterFunction4QueryBuilder.toArray(new FilterFunctionBuilder[0]))
                 .boostMode(CombineFunction.MULTIPLY).scoreMode(ScoreMode.MULTIPLY);
-
+        
         // @formatter:off
         m_queryBuilderForTopLevelFilter = QueryBuilders.boolQuery()
                 .should(QueryBuilders.boolQuery().mustNot(QueryBuilders.existsQuery("housenumber")))
                 .should(QueryBuilders.matchQuery("housenumber", query).analyzer("standard"))
-                .should(QueryBuilders.existsQuery(String.format("name.%s.raw", language)));
```
simonpoole, line 95: Removing this would seem to partially break querying for objects with names. I added this back and the behaviour seems to be much better.

DavidGorritiGeograma: Thank you! I will do more tests as well.

juanpelegrina:
Dear @simonpoole, 
I would like to know the status of this pull request. Are you planning to accept it? we have been using the code and we have not got any errors.
thanks for your help.
Cheers
Juan

simonpoole:
> 
> 
> Dear @simonpoole,
> I would like to know the status of this pull request. Are you planning to accept it? we have been using the code and we have not got any errors.
> thanks for your help.
> Cheers
> Juan

Hi Juan

I'm not a maintainer of this repo and wont be able to help directly. @lonvia any input?

lonvia:
I need to find time to review this. Not sure when this will happen.

lonvia:
Do you mind rebasing it on the latest master and force-push then new version? That would help.

DavidGorritiGeograma:
> Do you mind rebasing it on the latest master and force-push then new version? That would help.

Hello @lonvia, we have rebased our code with the latest fix for the test.  
Thanks a lot for your help!
David

src/main/java/de/komoot/photon/query/.gitignore
```diff
@@ -0,0 +1 @@
+/PhotonQueryBuilderBackup.java
```
otbutz: Putting java files in gitignore seems wrong

DavidGorritiGeograma: Hello, I deleted the .gitignore file. 
Thank you very much!
David

juanpelegrina:
Hi @lonvia, 
Could be possible to accept this pull request?
Regards
Juan

DavidGorritiGeograma:
Dear @lonvia,
 
We have been working locally with the solution which we developed in this PR, and we can confirm the results improve. We have been testing our proposal and we would like to share with you the results. Please check on this map the results of our testing
 
https://www.google.com/maps/d/viewer?mid=1JSQEeEBy0VyQRB2BOjo3RgNnXO1N4vF1&ll=54.45751685160265%2C15.605748222341717&z=4
 
Thanks for your comment and advice!
David

lonvia:
I have looked into this over Christmas. As far as I can see, the flaky housenumbers come from a different issue. We don't handle mixed-ligual searches very well Pretty much what #496 describes. Housenumbers are always in the original name while other parts tend to come from the language-specific parts. That's why they get hit by this more often.

I've looked through some of the examples from your Google document and as far as I can see the problem with them is that other parts of the address are missing in the database. The second round in your PR is more lenient with that.

I think that your approach with a second round has merit but I'd like to see the mixed-lingual problem solved first. I've experimented with multi-match on [this branch](https://github.com/lonvia/photon/tree/multi-match-all-languages). It fixed pretty much all the bug reports except for #227. And I still have to track down what the issue with that is.

lonvia:
Last years changes to the search algorithm have vastly improved the behaviour wrt housenumbers. So lcosing here.

PR #398, opened 2019-04-04
CORS added for cross-site requests and readme updated

simonpoole:
This is based on https://github.com/komoot/photon/pull/373 and preserves rahulhaque commits but simplifies things a tiny bit and rebases it on the current master.

Currently the -cors flag is all or nothing, but I would expect that people running photon that want more fine grained control would put it behind a reverse proxy in any case.

It doesn't resolve the mystery of why the previous code that added a Access-Control-Allow-Origin header seems to have worked for some (for example myself) and not for others.


src/main/java/de/komoot/photon/App.java
```diff
@@ -134,9 +135,17 @@ private static void startNominatimImport(CommandLineArgs args, Server esServer,
      * @param esNodeClient
      */
     private static void startApi(CommandLineArgs args, Client esNodeClient) {
-        setPort(args.getListenPort());
-        setIpAddress(args.getListenIp());
-
+        port(args.getListenPort());
+        ipAddress(args.getListenIp());
```
lonvia, line 142: This looks unrelated. Any particular reason to change it?

lonvia:
I suspect that the difference if the old version works or not is if the caller does an OPTION request first.

src/main/java/de/komoot/photon/App.java
simonpoole: The old versions are deprecated in spark but functionally exactly the same, changed here to get rid of the warnings.

I'll create a separate PR if you promise to merge it :-)

simonpoole:
As a tendency I would prefer -cors-all without an argument and -cors (or -cors-origin) with one.

simonpoole:
> 
> 
> I suspect that the difference if the old version works or not is if the caller does an OPTION request first.

I assume so too, I just couldn't recreate with any of the (very mainstream and current) browsers I use.

karussell:
This looks good. (Although I would not expose photon to the public and instead use nginx or similar tools to proxy it for unrelated security reasons)

prasadrajuv:
add https (ssl) support to avoid Blocked loading mixed active content in browser because geolocation support removed for http protocol

PR #399, opened 2019-04-07
Copy dependencies required for painless before resource processing

simonpoole:
This removes the jar files required for the painless plugin from the
repo and copies them before resources are processed.

This doesn't fix the ugliness in Server.java when coping the jars,
to make that at least a bit better and nearly only dependent on
configuration we strip the version when copying.

Further it converts indenting in pom.xml to spaces just as the rest of
the code.


pom.xml
```diff
@@ -1,245 +1,293 @@
...
+                <version>3.1</version>
+                <configuration>
+                    <encoding>UTF-8</encoding>
+                </configuration>
+            </plugin>
+        </plugins>
+        <resources>
+            <resource>
+                <directory>${project.build.directory}/es/</directory>
+            </resource>
```
lonvia, line 270: Those three lines were added and look unrelated. Or rather, if anything, they look like they should be replacing the three lines below

lonvia:
Looks in general sensible to me but my knowledge of Java dependency management is limited. @karussell do you know of any reason why we should keep those three packages around as jars in the git repo?

pom.xml
simonpoole: No that's correct, that is the temp directory the jars are copied in to (see line 229 for example) which then in turn is used by the resources plugin as the source for creating the jars and classes directories.

simonpoole:
@lonvia maybe I should have pointed this out: the jar files in question are a runtime dependency and are used for an ES plugin that provides some scripting support. maven (even though I hate it) handles compile time dependencies just fine, but this was a tiny bit tricky.

PR #400, opened 2019-04-07
Add some tests of the API

simonpoole:
This adds some integration tests for the API (in a wide sense of the word), further I've added the jacoco plugin so that we can at least get local coverage reports (integrating with travis is left as an exercise for the reader :-)).The test class is currently intermingled with the unit tests and run by the surefire plugin and not by failsafe, I don't think separation makes at lot of sense at this stage.

Note that this is based on https://github.com/komoot/photon/pull/399 and will need to be rebased before merging one way or the other.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -273,6 +273,11 @@
                 <groupId>org.jacoco</groupId>
                 <artifactId>jacoco-maven-plugin</artifactId>
                 <version>0.8.3</version>
+                <configuration>
+                    <excludes>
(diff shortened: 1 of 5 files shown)


lonvia:
No general objection to this but please remove the DFS_QUERY_THEN_FETCH commit. Make sure the test databases get created with one shard instead.

simonpoole:
@lonvia done (it was "slightly" more work than it seems, as all the elegant solutions don't actually work).

karussell:
See #420

Issue #396, opened 2019-03-28
Field city missing for addresses in Vienna


dthulke:
Responses to queries for addresses inside Vienna, Austria do not contain the city attribute. For example http://photon.komoot.de/api/?q=Stephansplatz%20Wien&lang=en .

Is this behaviour intended, due to a mapping error or a bug in Photon?
In contrast, results for other city states always contain both the city and state field (e.g. Hamburg and Salzburg) or only contain the city but note the state field (e.g. Berlin).

simonpoole:
Where the bug is a matter of debate, but photon currently has a hardwired list of these special cases which doesn't contain Vienna. If all goes well I should be able to test if adding Vienna to that list actually fixes the issue and have a look at at least making the list end-user configurable.

lonvia:
#457 will fix this.

lonvia:
This is now fixed in the extracts provided by graphhopper. I expect that the komoot instance gets updated soon.

PR #401, opened 2019-04-09
Externalize curated cities list

simonpoole:
This moves the list of curated cities list to a file and adds a small test that reading it actually works.

Further it adds the boundary relations for

-Algiers
-Moscow
-Vienna
-Havanna

to the list, this is pruned down from a list of 12 capitals that could have been affected by the issue. Additional verification that this is reasonable for the four cities would be helpful (Vienna should be OK, see https://github.com/komoot/photon/issues/396). Taipei might be a further candidate, but Photon seems to have issues finding anything there.


lonvia:
I have to say that I would prefer getting rid of the list completely. For the German cities currently hard-coded, looking at `place:de` would catch most of them. The missing ones have a `admin_title` = `Kreisfreie Stadt`.

Eventually this problem needs to be solved in Nominatim. It should report the cities correctly.

simonpoole:
Oh sure, don't disagree. This is a short term fix to some important cities being broken, and making it tiny bit easier to update the list as long as we don't have an alternative.

As I mentioned I looked a bit into any other way this could be determined completely on tagging alone, I'm not quite sure that exists. But as you say even if we can't completely do without an ugly curated list it should really be part of fixing the classification in Nominatim and not Photon.

lonvia:
So we have discussed this a bit. If you move it into an external list, then the list should also be configurable at runtime not just compile time. However, adding this is really too much hassle given that this should hopefully be fixed in Nominatim soon. I'll make a PR to add the four cities in the existing list for the short term fix.

lonvia:
#422 now contains a trimmed down version. I've kept the test. Closing here.

PR #403, opened 2019-04-10
Use SearchType.DFS_QUERY_THEN_FETCH for consistent scoring

simonpoole:
While working on adding a test with location bias in https://github.com/komoot/photon/pull/400 I was getting back a best scored result that was further away from the queried location than other (not really different) documents. 

I investigated and this behaviour (widely varying scores) remained even when I used -identical- documents with the exception of the ids and wasn't using location bias. The difference in scores was roughly a factor two. This isn't good because lots of scoring in Photon multiplies these values with other scores making the erroneous results worse (for example when using location bias). 

As far as I could determine (aka half a day later) this is simply an artefact of the default scoring mechanism, and it goes away when using SearchType.DFS_QUERY_THEN_FETCH , however I wouldn't rule out that the behaviour is caused by the small number of documents in the index during the test (18). It would be really good if somebody with more ES knowledge could weigh in.

### DIFF
--- a/src/main/java/de/komoot/photon/searcher/BaseElasticsearchSearcher.java
+++ b/src/main/java/de/komoot/photon/searcher/BaseElasticsearchSearcher.java
@@ -21,7 +21,7 @@ public class BaseElasticsearchSearcher implements ElasticsearchSearcher {
     public SearchResponse search(QueryBuilder queryBuilder, Integer limit) {
         TimeValue timeout = TimeValue.timeValueSeconds(7);
         return client.prepareSearch("photon").
-                setSearchType(SearchType.QUERY_AND_FETCH).
+                setSearchType(SearchType.DFS_QUERY_THEN_FETCH).
                 setQuery(queryBuilder).
                 setSize(limit).
                 setTimeout(timeout).


simonpoole:
Results from running the geocoder tests

```
      Failed Passed Time
0.3.1 15'451  6'733  660s
this  15'441  6'743  970s (40 new failed, 50 new passed)
```

hbruch:
Thanks for evaluating this change and providing the test results!

I guess that the widely varying score is indeed caused by the small number of documents. Elastic's article ["Understanding Query-then-fetch vs DFS-Query-then-fetch"](https://www.elastic.co/de/blog/understanding-query-then-fetch-vs-dfs-query-then-fetch) explains how varying scores might be caused by different TF-IDF scores, depending on the shard a document is retrieved from. I suppose that with greater number of addresses, the variance will be far less.

However, as described in the article, DSF-Query-Then-Fetch results in an additional round-trip which might explain the 970s accumulated response time with DSF-QTF vs 660s with QTF you experienced in your tests(?)

50 new tests passed, but as 40 new tests failed and the change seems to have a performance impact, I'd rather not switch to DFS-QTF.

simonpoole:
While I would agree that taking the performance hit just on theoretical grounds and because I want to write a reproducible test, couldn't be justified, I can reproduce the behaviour on an index for the planet.

Consider the following query

` ?q=Switzerland&osm_tag=amenity:fast_food&limit=5&lat=47.41840465&lon=8.3739254`

the coordinates are bang on the local McDonald's. With the default search strategy the top result is a restaurant half a km away, with DFS_QUERY_THEN_FETCH the ordering is correct with the original top result at position three. Naturally this is particularly visible because we multiple the scores with the result of the location bias and objects that are close together will not have massively different biases.

My gut feeling is while we perhaps don't want to use DFS_QUERY_THEN_FETCH all the time, we could:
- make it configuration/flag dependent
and/or
- use it by default for queries with location bias

lonvia:
I agree with @hbruch here that this is not a good idea (DFS_QUERY_THEN_FETCH essentially runs each query twice). The ES book suggests to run the tests with only one shard to make the problem with the IDF go away.

simonpoole:
@lonvia see my example it happens irl too, while for testing it is simply annoying and not really important, I don't think it can be easily communicated to end users that they are getting wrong results, but they should still be happy because they get the wrong answers faster.

I would suggest deferring the decision on this till I've benchmarked this on location biased queries only (the overwhelming bulk of the geo-tester tests are without location).

lonvia:
Just because it happens to fix one example doesn't mean that it is the correct fix. If DFS is needed then there is an underlying problem with the index we should be fixing. Please provide the full scoring log for your example for queries with and without DFS. Maybe there is something to see there.

lonvia:
Giving it some more thought, the problem with your example (and also the ones with Starbucks and McDonalds) is that the query terms themselves end up with a very low scoring because of the low IFS (i.e. they are very frequent terms). We might want to investigate if the scoring needs to be improved for these cases.

simonpoole:
That could well be (iirc the scores were around 50).

lonvia:
Closing for now. If you want to pursue the problems with the given examples further, please open a ticket where we can discuss different options.

Issue #183, opened 2015-08-25
Provide countryCode and stateCode in response


vmeurisse:
Today the json returned contain the country name and state name but not the codes.

It would be really nice to have the codes as well as a lot of services using addresses require them.
This would also help for #175 as the user could simply postfilter the result for a given country.

Aleksio:
This would be a great addition and would cover my issue completely!
I tried looking into fixing it myself (even though my knowledge of Java is really limited), but it seems that the dataset in elasticsearch doesn't have the data either (correct me if I'm wrong), so I think the data needs to be added in the Nomatim import?

I guess we could do that ourselves, but then we can't profit from the ready to use search index provided by graphhopper.

vpena:
Hello!

Any news about this? Is it possible to get the country code in some way?

Thank you very much,
Vanessa

karussell:
We love to receive pull requests :) !

vpena:
Hehe, I would love to help but I probably won't find the time soon :P However, if it is useful to anyone, I found these libraries that might help:

https://pypi.python.org/pypi/pycountry
https://github.com/tuneinc/pycountry-convert

I use them as:

import pycountry_convert
country_name = result_raw['properties']['country']
country_code = pycountry_convert.country_name_to_country_alpha2(country_name, cn_name_format="default")

Cheers and thanks for your answer!
Vanessa

PR #404, opened 2019-04-11
Add country code to the response

Azbesciak:
To keep convention field is named 'countrycode' like a 'housenumber'. It is also uppercase (ISO-3166 alpha 2).
It also partly fulfills https://github.com/komoot/photon/issues/183.


lonvia:
I don't have any objection to add the country code to the database and returning it with the result. I'm less sure about adding it to the collector to make it searchable. I'd expect two letter words to interfere in a bad way there.

otbutz:
> I don't have any objection to add the country code to the database and returning it with the result. I'm less sure about adding it to the collector to make it searchable. I'd expect two letter words to interfere in a bad way there.

Being able to search for something like `Paris US` would be pretty nice.

lonvia:
Superseded by #425.

PR #409, opened 2019-04-14
Update dependencies for Java 11

simonpoole:
Updated dependencies from @lonvia The specific mockito version has transitive dependencies that cause issues and one of which needed to be excluded.


lonvia:
Bumped lombok to the current 1.18.6. If CI likes it then there is no reason to not go to the latest version.

PR #410, opened 2019-04-14
Use getUid when updating not getPlaceId

simonpoole:
This is likely a bug that causes duplication of house number documents in the index and potentially overwriting of  documents . I'm in the process of determining if that is actually the case. IMHO this should in the end only be a serious (as in destructive overwrite) issue for address interpolations that have been expanded in photon.


simonpoole:
This has been superseded by https://github.com/komoot/photon/pull/412

PR #412, opened 2019-04-20
Update fix

simonpoole:
-don't expand multiple housenumbers on the same object
-use a nominatim instance independent document id
-delete documents expanded from interpolations by the osm data

This fixes the remaining update problems outside of https://github.com/openstreetmap/Nominatim/issues/1360 It currently continues to expand address interpolations and in general is rather ugly. As the mapping has changed this requires a re-import.

The alternative is to not expand interpolations, use a distance based scoring to the actual interpolation way and generate individual results from that on the fly. This will require changes to the queries and will have an unknown performance impact.

### DIFF
--- a/src/main/java/de/komoot/photon/PhotonDoc.java
+++ b/src/main/java/de/komoot/photon/PhotonDoc.java
@@ -7,8 +7,12 @@ import lombok.Getter;
 import lombok.Setter;
 
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 
 /**
  * denormalized doc with all information needed be dumped to elasticsearch
@@ -38,10 +42,11 @@ public class PhotonDoc {
     private Set<Map<String, String>> context = new HashSet<Map<String, String>>();
     private Map<String, String> country;
     private Map<String, String> state;
-    private String houseNumber;
+    private List<String> houseNumbers;
     private Point centroid;
+    private boolean isExpanded = false;
 
-    public PhotonDoc(long placeId, String osmType, long osmId, String tagKey, String tagValue, Map<String, String> name, String houseNumber, Map<String, String> extratags, Envelope bbox, long parentPlaceId, double importance, CountryCode countryCode, Point centroid, long linkedPlaceId, int rankSearch) {
+    public PhotonDoc(long placeId, String osmType, long osmId, String tagKey, String tagValue, Map<String, String> name, @Nullable List<String> houseNumbers, Map<String, String> extratags, Envelope bbox, long parentPlaceId, double importance, CountryCode countryCode, Point centroid, long linkedPlaceId, int rankSearch) {
         String place = extratags != null ? extratags.get("place") : null;
         if (place != null) {
             // take more specific extra tag information
@@ -55,7 +60,7 @@ public class PhotonDoc {
         this.tagKey = tagKey;
         this.tagValue = tagValue;
         this.name = name;
-        this.houseNumber = houseNumber;
+        this.houseNumbers = houseNumbers;
         this.extratags = extratags;
         this.bbox = bbox;
         this.parentPlaceId = parentPlaceId;
@@ -73,7 +78,7 @@ public class PhotonDoc {
         this.tagKey = other.tagKey;
         this.tagValue = other.tagValue;
         this.name = other.name;
-        this.houseNumber = other.houseNumber;
+        this.houseNumbers = other.houseNumbers;
         this.postcode = other.postcode;
         this.extratags = other.extratags;
         this.bbox = other.bbox;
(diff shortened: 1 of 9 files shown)


src/main/java/de/komoot/photon/nominatim/NominatimUpdater.java
```diff
@@ -113,12 +109,11 @@ public void update() {
                         template.update("update location_property_osmline set indexed_status = 0 where place_id = ?;", placeId);
 
                         Integer indexedStatus = place.getIndexdStatus();
-                        if (indexedStatus != CREATE) {
-                            updater.delete(placeId);
-                            if (indexedStatus == DELETE) {
-                                deletedInterpolations++;
-                                continue;
-                            }
+                        // nominatim is rather flaky about setting the update flag so lets always delete
```
lonvia, line 112: Sounds like something that needs fixing in Nominatim. Can you be more specific?

src/main/java/de/komoot/photon/nominatim/model/UpdateRow.java
```diff
@@ -10,4 +10,8 @@
 
     public Long placeId;
     public Integer indexdStatus; // 1 - index, 2 - update, 100 - delete
+    private String osmType;
+    private long osmId;
+    private String osmKey;
+    private String osmValue;
```
lonvia, line 16: Doesn't look like osmValue is used anywhere.

src/main/java/de/komoot/photon/nominatim/NominatimUpdater.java
```diff
@@ -57,13 +57,10 @@ public void update() {
...
                             Integer indexedStatus = place.getIndexdStatus();
-                            if (indexedStatus == DELETE || (indexedStatus == UPDATE && rank == MAX_RANK)) {
-                                updater.delete(placeId);
-                                if (indexedStatus == DELETE) {
-                                    deletedPlaces++;
-                                    continue;
-                                }
-                                indexedStatus = CREATE; // always create
+                            if (indexedStatus == DELETE) {
+                                updater.delete(PhotonDoc.getBaseId(place.getOsmType(), place.getOsmId(), place.getOsmKey()));
```
lonvia, line 61: Please add a getUid function to updateDoc.

src/main/java/de/komoot/photon/elasticsearch/Server.java
```diff
@@ -107,6 +108,7 @@ public Server start() {
                 Settings settings = sBuilder.build();
                 Collection<Class<? extends Plugin>> lList = new LinkedList<>();
                 lList.add(Netty4Plugin.class);
+                lList.add(ReindexPlugin.class);
```
lonvia, line 111: What is that for?

src/main/java/de/komoot/photon/PhotonDoc.java
```diff
@@ -88,27 +93,50 @@ public PhotonDoc(PhotonDoc other) {
...
-        if (houseNumber == null || houseNumber.isEmpty())
-            return String.valueOf(placeId);
-        else
-            return String.valueOf(placeId) + "." + houseNumber;
+        String id = getBaseId(osmType, osmId, tagKey);
+        if (houseNumbers == null || houseNumbers.isEmpty()) {
+            return id;
+        } else {
+            return id + (isExpanded ? "." + houseNumbers.get(0) : "");
+        }
```
lonvia, line 111: Always adding the housenumber looks wrong now with respect to the updates below. You don't add the housenumbers there (and can't because you don't have the old housenumber). Given that simple house number objects are not expanded anymore, you should only add the housenumber to interpolation documents. Then the updates work with using the simple base id.

simonpoole:
To explain this a bit better, these changes fix the update issues without changing anything wrt scoring and querying, as they still require changing the id generation and a re-import, it doesn't seem to make sense to not change the id generation in this PR, particularly considering that it isn't clear if a solution with non-expanded interpolations will actually work satisfactory (I'm looking at that, but IMHO it is a separate undertaking). More comments inline.

src/main/java/de/komoot/photon/PhotonDoc.java
simonpoole: isExpanded is only set to true for interpolations.

src/main/java/de/komoot/photon/elasticsearch/Server.java
simonpoole: Required for the remote delete query to work see https://github.com/komoot/photon/pull/412/files#diff-0ed928df648038beadef62e9712e4397 line 82

src/main/java/de/komoot/photon/nominatim/NominatimUpdater.java
simonpoole: I don't quite understand that. We don't have a PhotonDoc instance at this point in the code, we could requery the database and generate one as per the normal flow, so that we could use its getUid method but it would seem to be unnecessarily complicated and in the end not do anything different.

src/main/java/de/komoot/photon/nominatim/NominatimUpdater.java
simonpoole: In certain circumstances, iirc adding a new end node to an interpolation, replacing the old one. Nominatim will set the CREATE flag and not UPDATE. I can re-test that locally and open a Nominatim issue,.

src/main/java/de/komoot/photon/nominatim/model/UpdateRow.java
simonpoole: I need to check, but likely both osmKey and osmValue aren't used any more (an initial implementation deleted non-interpolation expanded documents this way).

lonvia:
I think I know how to fix the update mess around expanded housenumbers and interpolations without any changes to the document ID. So finally closing this one.

PR #413, opened 2019-05-07
Add respecting Accept-Language header

Azbesciak:
Will be applied as a fallback in case of missing `lang` query param, only if supported languages intersects languages in the header with header priority spec maintained.

### DIFF
--- a/src/main/java/de/komoot/photon/query/LanguageChecker.java
+++ b/src/main/java/de/komoot/photon/query/LanguageChecker.java
@@ -19,9 +19,13 @@ public class LanguageChecker implements Function<String, Boolean, BadRequestExce
     @Override
     public Boolean apply(String lang) throws BadRequestException {
         if (lang == null) lang = "en";
-        if (!supportedLanguages.contains(lang)) {
+        if (!isLanguageSupported(lang)) {
             throw new BadRequestException(400, "language " + lang + " is not supported, supported languages are: " + Joiner.on(", ").join(supportedLanguages));
         }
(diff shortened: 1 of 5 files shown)


Azbesciak:
Any feedback?

lonvia:
Sorry for the late review. So, I think this is useful in general. I have three comments about the implementation though:

1. It needs to be implemented for the ReverseRequestFactory as well.
2. What about the language strings like 'en-US', which are common in Accept-Language headers?
3. I find wrapping the LanguageChecker class a little bit overkill. I would just merge the new RequestLanguageResolver and LanguageChecker in one.

otbutz:
>     2. What about the language strings like 'en-US', which are common in Accept-Language headers?

Works flawlessly using [Locale.lookupTag()](https://docs.oracle.com/javase/8/docs/api/java/util/Locale.html#lookupTag-java.util.List-java.util.Collection-)

Minimal working example:
```java
List<String> supportedLangs = Arrays.asList("de", "en", "es", "fr");
String ranges = "fr-CH, fr-FR;q=0.9, en;q=0.8, de;q=0.7, *;q=0.5";
List<Locale.LanguageRange> languageRanges = Locale.LanguageRange.parse(ranges);
String tag = Locale.lookupTag(languageRanges, supportedLangs);
System.out.println(tag); // returns "fr"
```

lonvia:
Ah, so we only need a test for it. :)

lonvia:
Closing in favour of #467.

PR #421, opened 2019-06-14
Bbox

Gerungofulus:
Extension of #364 by @hbruch:
I basically just rebased origin/master as all previous remarks had already been fixed by hbruch

### DIFF
--- a/src/main/java/de/komoot/photon/query/BoundingBoxParamConverter.java
+++ b/src/main/java/de/komoot/photon/query/BoundingBoxParamConverter.java
@@ -26,8 +26,8 @@ public class BoundingBoxParamConverter implements Function<Request, Envelope, Ba
                 throw new BadRequestException(400, INVALID_BBOX_ERROR_MESSAGE);
             }
             Double minLon = Double.valueOf(bboxCoords[0]);
-            Double minLat = Double.valueOf(bboxCoords[2]);
-            Double maxLon = Double.valueOf(bboxCoords[1]);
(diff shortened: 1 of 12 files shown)


src/main/java/de/komoot/photon/query/BoundingBoxParamConverter.java
```diff
@@ -0,0 +1,45 @@
...
+            Double minLat = Double.valueOf(bboxCoords[2]);
+            Double maxLon = Double.valueOf(bboxCoords[1]);
+            Double maxLat = Double.valueOf(bboxCoords[3]);
+            if (minLon > 180.0 || minLon < -180.00 || maxLon > 180.0 || maxLon < -180.00) {
+                throw new BadRequestException(400, INVALID_BBOX_ERROR_MESSAGE);
+            }
+            if (minLat > 90.0 || minLat < -90.00 || maxLat > 90.0 || maxLat < -90.00) {
+                throw new BadRequestException(400, INVALID_BBOX_ERROR_MESSAGE);
+            }
+            bbox = new Envelope(minLon, minLat, maxLon, maxLat);
```
lonvia: There is a double error in here that happens to cancel itself out. bbox above is parsed as (minLon, maxLon, minLat, maxLat) while it should be (minLon, minLat, MaxLon, MaxLat). In the constructor for Envelope the exact reverse is true.

Gerungofulus: Changed code according to the documentation

PR #424, opened 2019-06-15
[WIP] Test nominatim import

karussell:
It would be great if we can test the import too. This should work now locally.  I have also roughly documented via the travis file how to get this up and running locally.

Further tasks:

 * make the travis build working
 * exclude this test in the normal "mvn install" run? (if we want this, I can have a look)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,3 +3,22 @@ language: java
 jdk:
   - oraclejdk8
   - oraclejdk9
+
+cache:
+  directories:
+    - $HOME/dump
+
+addons:
+  postgresql: "10"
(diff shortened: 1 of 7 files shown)


lonvia:
With the new H2 test harness we have a way of unit testing the Nominatim importer now. I would suggest we close this PR and instead set up a full "Setup Nominatim -> Import into Nominatim -> export to Photon -> query Photon" github action at some point that makes sure we remain compatible with the actual Nominatim and also test the Postgres side of things.

lonvia:
CI with full import is now working. Closing this.

Issue #418, opened 2019-06-13
feature request: country bias


alexandervlpl:
It would be great to increase the priority of all results from a certain country.

I'm trying to achieve this using location bias. I get a noticable effect with a `location_bias_scale` of 4 or higher, but then even insignificant features very close to the location get priority and important results farther away but in the same country are left out. :cry:

For example, using a point in Kansas as a "center" for the United States, I search "new". I get many small towns in Kansas but no "New York":

`/api?q=new&limit=5&lat=39&lon=-98&location_bias_scale=4`

lonvia:
We can have a simple country filter. That is fairly easy.

alexandervlpl:
Ideally, I think there should be two distinct country parameters like in the [HERE geocode API](https://developer.here.com/documentation/geocoder/topics/resource-geocode.html):

- **country**: restict results to one or more countries (I see a few issues about this already).
- **countryfocus**: _prioritize_ results from one or more countries, but still include global results. Similar to location bias. This is what I meant.

lonvia:
Well, a filter is easy. Reranking is a bit more tricky.

wottpal:
Having a country filter but be great!

PR #425, opened 2019-06-16
Country codes

Gerungofulus:
This PR builds on #404. It is different in the sense that it doesn't not add country codes to the default collector. It partially implements #183, as mentioned by the original contributor @Azbesciak. I originally set out to add a country code query param as suggested by @lonvia in #418, but I am unsure about the best way to implement it. Photon already implements filters by tags by using the query parameters `osm_tag`. I think filtering addresses should be possible by filtering for an osm tag. Unfortunately I was not able to filter properly by country codes using this method. The other solution would be to simply add a new query parameter `country_code`.  What do you think?


Gerungofulus:
Okay I just realised that osm_tag does not behave in the sense that it could filter any key value pair as it would be possible working directly on osm postgre data. We should probably rename that filter to avoid confusion in the future. What do you think?

karussell:
Looks good to me!

And yes, the filtering should go into a different PR. What exactly did you find confusing about osm_tag?

Gerungofulus:
> Looks good to me!
> 
> And yes, the filtering should go into a different PR. What exactly did you find confusing about osm_tag?
Well as far as I remember, a given node or way in OSM is an object with an ID in the corresponding table in (nodes/ways). Those objects can have an indefinite amount of key-value pairs called tags. The naming scheme of photon implies that only a certain selection of these kv-pairs is indeed a tag (and an object may only correspond to one of these tags?). So that's why I find the naming osm_tag relatively confusing, as kv pairs with the key being addr:street would not be considered tags while they indeed are considered tags at other sources such as [this](https://taginfo.openstreetmap.org/projects/nominatim#tags).

karussell:
> The naming scheme of photon implies that only a certain selection of these kv-pairs is indeed a tag

I think this is only done because indexing them can be expensive (in terms of index size). Indexing all tags is not easy for world wide setup, so we index only an "important" subset

Gerungofulus:
> > The naming scheme of photon implies that only a certain selection of these kv-pairs is indeed a tag
> 
> I think this is only done because indexing them can be expensive (in terms of index size). Indexing all tags is not easy for world wide setup, so we index only an "important" subset

Yeah that's what I was thinking. It's just that I suggest renaming the query parameter or at least listing which tags are compatible, because this [list](https://taginfo.openstreetmap.org/projects/nominatim#tags) which is linked in the Readme, does contain country codes, but adding this key to the query params via `osm_tag` does not seem to do much.

karussell:
Yes, we should probably throw an exception for tags that we currently do not support and list all supported (similarly what we do with the query params)

lonvia:
Country code filtering really should be a separate parameter. It is independent of tag filtering.

PR #454, opened 2020-04-06
Add command line option to fetch updates and exit afterwards

Nakaner:
When doing Photon upgrades on a (master) host not serving requests but getting its database rsynced to remote hosts, it is pretty annoying that it is only visible in the logfile when an update is completed. Therefore, I propose to add a new option `-nominatim-update` which makes Photon start, apply updates from Nominatim and then exit.

### DIFF
--- a/src/main/java/de/komoot/photon/App.java
+++ b/src/main/java/de/komoot/photon/App.java
@@ -64,8 +64,17 @@ public class App {
                 return;
             }
 
-            // no special action specified -> normal mode: start search API
-            startApi(args, esClient);
+            if (args.isNominatimUpdate()) {
+                shutdownES = true;
+                log.info("Ensuring that the cluster is ready, this might take some time.");
+                // inspired by https://stackoverflow.com/a/50316299
+                esClient.admin().cluster().prepareHealth().setWaitForGreenStatus().execute().actionGet();
+                final NominatimUpdater nominatimUpdater = setupNominatimUpdater(args, esClient);
+                nominatimUpdater.update();
+            } else {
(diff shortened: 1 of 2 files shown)


src/main/java/de/komoot/photon/CommandLineArgs.java
```diff
@@ -20,6 +20,9 @@
     @Parameter(names = "-nominatim-import", description = "import nominatim database into photon (this will delete previous index)")
     private boolean nominatimImport = false;
 
+    @Parameter(names = "-nominatim-update", description = "fetch updates from nominatim database into photon and exit (this updates the index only without offering an API")
```
lonvia, line 23: Missing closing bracket in the description.

src/main/java/de/komoot/photon/App.java
```diff
@@ -64,8 +64,17 @@ public static void main(String[] rawArgs) throws Exception {
...
-            // no special action specified -> normal mode: start search API
-            startApi(args, esClient);
+            if (args.isNominatimUpdate()) {
+                shutdownES = true;
+                log.info("Ensuring that the cluster is ready, this might take some time.");
+                // inspired by https://stackoverflow.com/a/50316299
+                esClient.admin().cluster().prepareHealth().setWaitForGreenStatus().execute().actionGet();
+                final NominatimUpdater nominatimUpdater = setupNominatimUpdater(args, esClient);
+                nominatimUpdater.update();
+            } else {
```
lonvia, line 74: Please keep with the style of the other ifs and do the `return;` instead of the `else`.

PR #461, opened 2020-05-23
Update database dependencies

lonvia:
This updates the postgresql and postgis packages as well as commons-dbcp and spring-jdbc. Hstore is now natively supported, so we can drop the special osmosis hstore package.


lonvia:
On  a positive note, these updated libraries make the import from Nominatim around 50% faster.

PR #468, opened 2020-06-10
Import and index county field

leonardehrenfried:
As requested in #303 I've added the county field and removed the duplication.


lonvia:
Excellent. Now we soon have to make a new release so that all the fancy new fields show up on photon.komoot.de as well.

PR #500, opened 2020-10-05
Convert website to simple static html

lonvia:
The website is simple enough that we don't really need flask or templating to serve it. Therefore remove Python completely from the project and convert the website to a simple index.html that can be served statically.

@christophlingg, please let me know if I have overseen something that requires the flask setup. This change removes a quiet large moving part for the website setup but also a couple of dependencies that regularly get us security warnings on the github repo.


website/config.js.example
```diff
@@ -0,0 +1,4 @@
+API_URL = 'http://localhost:5001/api/?';
+TILELAYER = 'https://{s}.tile.hosted.thunderforest.com/komoot-2/{z}/{x}/{y}.png';
```
christophlingg, line 2: I guess most people have high definition we could use `https://{s}.tile.hosted.thunderforest.com/komoot-2/{z}/{x}/{y}@2x.png`; instead. But maybe you also know about an option in leaflet to automatically change the URL depending on the display density? I found [detectReting](https://leafletjs.com/reference-1.7.1.html#tilelayer-detectretina) but that seems to produce pretty small maps.

christophlingg (approved):
It makes a lot of sense, when we started flask also served api request. but of course the leftovers now do not deserve a dynamic server.

website/config.js.example
lonvia: I'll have a look. I wanted to update leaflet anyways. That's a good opportunity to check that as well.

mtmail: Use `{r}` and leaflet will add `@2x` for those displays which support high resolution https://leafletjs.com/reference-1.7.1.html

PR #517, opened 2020-11-24
Introduce enum for Photon's address type

### LINKED ISSUE #499: Strange results for Dutch provinces
Using the 0.3.4 release and the Graphhopper database from yesterday, I'm getting strange and inconsistent results for Dutch provinces. http://localhost:2322/api?q=friesland gives me:

```
        "country": "The Netherlands",
        "osm_key": "boundary",
        "osm_value": "administrative",
        "name": "Friesland",
        "county": "Netherlands",
        "type": "region"
```

The Netherlands are returned as both country and county. Friesland is a province. In the operation version at http://photon.komoot.de/api/?q=friesland, a simpler but correct result is returned:

```
        "country": "The Netherlands",
        "osm_key": "place",
        "osm_value": "state",
        "name": "Friesland"
```

When running http://localhost:2322/api?q=leeuwarden (Leeuwarden being a city in Friesland):

```
        "country": "The Netherlands",
        "osm_key": "boundary",
        "city": "Leeuwarden",
        "osm_value": "administrative",
        "name": "Leeuwarden",
        "county": "Friesland",
        "state": "Netherlands",
        "type": "locality"
```

Here, Friesland is considered the county, rather than the Netherlands. The result at http://photon.komoot.de/api/?q=leeuwarden is again correct:

```
        "country": "The Netherlands",
        "osm_key": "place",
        "osm_value": "city",
        "name": "Leeuwarden",
        "state": "Friesland"
```

I assume this is a regression of some sort. Which version is currently running at photon.komoot.de?

lonvia:
The enum lists the possible address fields that Photon may index and output. It is also responsible for mapping Nominatim's address rank values to the address type. This should now be the only place where raw Nominatim address rank numbers are used.

PhotonDoc uses the type to define its address in a map instead of having lots of separate fields. That simplifies code in a number
of places.

The commit also disallows adding an address part with the same level as the place object itself. CC @hendrikmoree 

Fixes #499.

### DIFF
--- a/src/main/java/de/komoot/photon/PhotonDoc.java
+++ b/src/main/java/de/komoot/photon/PhotonDoc.java
@@ -4,8 +4,8 @@ import com.google.common.collect.ImmutableMap;
 import com.neovisionaries.i18n.CountryCode;
 import com.vividsolutions.jts.geom.Envelope;
 import com.vividsolutions.jts.geom.Point;
+import de.komoot.photon.nominatim.model.AddressType;
 import lombok.Getter;
-import lombok.Setter;
 import lombok.extern.slf4j.Slf4j;
 
 import java.util.*;
@@ -16,7 +16,6 @@ import java.util.*;
  * @author christoph
  */
 @Getter
-@Setter
 @Slf4j
 public class PhotonDoc {
     final private long placeId;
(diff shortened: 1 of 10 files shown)


src/main/java/de/komoot/photon/PhotonDoc.java
```diff
@@ -173,31 +149,54 @@ public void completeFromAddress() {
...
      *
-     * @param existingField The current value of the document's address field.
+     * @param addressType The type of address field to fill.
      * @param addressFieldName The name of the address tag to use (without the 'addr:' prefix).
      *
      * @return 'existingField' potentially with the name field replaced. If existingField was null and
      *         the address field could be found, then a new map with the address as single entry is returned.
      */
-    private Map<String, String> extractAddress(Map<String, String> existingField, String addressFieldName) {
+    private void extractAddress(AddressType addressType, String addressFieldName) {
```
karussell, line 158: Could you get addressFieldName also from addressType.getName? Then you could remove it from the parameter list.

BTW: thanks for introducing me to EnumMap - really interesting :)

lonvia: The mapping between the GeocodeJSON names and the OSM ` addr:*`  tag names is not exactly 1:1. suburb and neighbourhood are already different and we might even add more variants (like 'province'  for 'state'). We could add the mapping into the new AddressType enum as an extra field though. Something for an extra PR.

Issue #514, opened 2020-11-16
Different results when querying with diferent language preferences


hallpower:
I've submitted following request via the browser: https://photon.komoot.io/api/?q=muscat&osm_tag=aeroway:aerodrome

This yields following result: 

```json
{
    "features": [
        {
            "geometry": {
                "coordinates": [
                    58.29247709617641,
                    23.5987194
                ],
                "type": "Point"
            },
            "properties": {
                "city": "Al Mawalih ash Shamaliyah",
                "country": "Oman",
                "countrycode": "OM",
                "extent": [
                    58.2477853,
                    23.6155156,
                    58.329042,
                    23.5833579
                ],
                "name": "Muscat International Airport",
                "osm_id": 20440333,
                "osm_key": "aeroway",
                "osm_type": "W",
                "osm_value": "aerodrome",
                "postcode": "1122",
                "state": "Muscat",
                "street": "Way 808",
                "type": "house"
            },
            "type": "Feature"
        },
        {
            "geometry": {
                "coordinates": [
                    -91.14591628475205,
                    41.3660709
                ],
                "type": "Point"
            },
            "properties": {
                "city": "Muscatine",
                "country": "United States of America",
                "countrycode": "US",
                "county": "Muscatine County",
                "extent": [
                    -91.1606258,
                    41.3753627,
                    -91.1283817,
                    41.3590896
                ],
                "name": "Muscatine Municipal Airport",
                "osm_id": 104449438,
                "osm_key": "aeroway",
                "osm_type": "W",
                "osm_value": "aerodrome",
                "postcode": "52749",
                "state": "Iowa",
                "type": "house"
            },
            "type": "Feature"
        },
        {
            "geometry": {
                "coordinates": [
                    -91.2198191,
                    41.586144
                ],
                "type": "Point"
            },
            "properties": {
                "country": "United States of America",
                "countrycode": "US",
                "county": "Muscatine County",
                "name": "Walker Field",
                "osm_id": 1042011245,
                "osm_key": "aeroway",
                "osm_type": "N",
                "osm_value": "aerodrome",
                "postcode": "52776",
                "state": "Iowa",
                "street": "Hayes Avenue",
                "type": "house"
            },
            "type": "Feature"
        }
    ],
    "type": "FeatureCollection"
}
```

When I run the same request via Postman (or cURL), I get a different result set: 
```bash
curl --location --request GET 'https://photon.komoot.io/api/?q=muscat&osm_tag=aeroway:aerodrome'
```

[...]

Issue #512, opened 2020-11-13
geocoding quality decreased since version 0.3.4 


choesterle:
Hi,

I've noticed some quality deterioration for geocoding between version 0.3.2 and 0.3.4.

In earlier versions, when executing
https://photon.komoot.io/api?q=Ausstellungsgasse%2011%20Wiener%20Neustadt%20%C3%96sterreich&lang=de
a result was returned that explicitly matched the search criteria. That means the result contained the properties
_street: Ausstellungsgasse, 
housenumber: 11, 
city: Gemeinde Wiener Neustadt
etc._
With the new version 0.3.4, we don't get such a result containing all search criteria. The closest match is

```
properties": {
"osm_id": 459950166,
"country": "Österreich",
"city": "Wiener Neustadt",
"countrycode": "AT",
"postcode": "2700",
"locality": "Musikantenviertel (Gartenstadt)",
"type": "house",
"osm_type": "N",
"osm_key": "amenity",
"housenumber": "5",
"street": "Ausstellungsgasse",
"osm_value": "kindergarten",
"name": "NÖ Landeskindergarten Ausstellungsgasse",
"state": "Niederösterreich"
}
```

but with a wrong housenumber _5_ instead of _11_

BR,
Christian

lonvia:
This particular case actually has a data error. [way 107802295](https://www.openstreetmap.org/way/107802295) has ` addr:city=2700`. But I don' t think this is what is causing the issues here. It looks like we have a problem with cross language lookups. This used to work. I need to find out what changed.

lonvia:
The example has been fixed in the meantime which fixed the query. #538 should also fix the underlying mixed-lingual issue.

Issue #496, opened 2020-09-25
Housenumber issue when searching in non-default language


kostjerry:
It would be helpful to add the **housenumber** field to the all lang-specific collectors.

For now the address search in other langs miss many addresses.
I have tested the address "Тростянецька 3" in Kyiv (Ukraine) with default lang **uk** and additional langs **ru** and **en**.
When requesting "Trostianetska 3" in **en**, the photon don't find this address.
I have solved this issue by adding all langs manually in mappings.json and rebuilding jar-file.
```
...
"housenumber": {
	"type": "text",
	"index": true,
	"analyzer": "index_housenumber",
	"search_analyzer": "standard",
	"copy_to": [
		"collector.default",
		"collector.uk",
		"collector.ru",
		"collector.en"
	]
},
...
```
I believe this could be done automatically from the code, but unfortunately I was not able to achieve the desired result by this way.

lonvia:
This is a really good catch but the issue is not confined to house numbers. Any mixing of languages in the query will go wrong. So I went for a more broad approach in #538. Could you give that a try and see if it resolves your issues without making other queries worse? The PR can be run on an existing database. It only changes the query to ES.

lonvia:
Searching the address in English now works on the demo server. So I assume this is fixed.

PR #538, opened 2021-02-12
Searching with mixed-lingual queries

lonvia:
Photon saves addresses in language-specific collectors and searches them separately when a query comes in. The result is that you cannot do mixed-language queries. That's queries like 'Rindermarkt, Munich'  where the street name is in one language and the city name in another. This hits us especially hard with house numbers as they are only saved in the default collector. So you cannot search for house number addresses in any than the local language.

This PR switches the basic query to a multi-match strategy over all collectors, so that you can have the components of your address in any supported language. Unfortunately there is no support for fuzzy matching with the cross-field multi-matcher. So the second lenient match falls back to the strategy we had now to search each collector separately.

In the long run it might be better to have one large address collector that simply collects all terms in all languages but for now the multi match can improve results without changing the database.

Fixes the examples given in #514, #512, #496, #227.

I've set up a demo server here: https://pagan.lonvia.de/ It would be nice if you could play with it and check if the multi-match results in any noticeable regressions. The data is the same as on https://komoot.photon.io.


lonvia:
This is now deployed on https://komoot.photon.io. Lets see how it works out there.

otbutz:
@lonvia https://photon.komoot.io/ ;)

lonvia:
Fixed.

Issue #529, opened 2021-01-21
Speed up import for rank 30 objects


lonvia:
When importing objects from Nominatim to Photon, the importer needs to read the address information from the `place_addressline` table for each object. For objects at rank 30 the information is not read for the object itself but from its parent object, see [NominatimConnector:179](https://github.com/komoot/photon/blob/master/src/main/java/de/komoot/photon/nominatim/NominatimConnector.java#L179)ff.

Import could be sped up quite a bit if the parent address was cached and reused between rank 30 objects with the same parent. I was thinking along the lines of sorting the [object query](https://github.com/komoot/photon/blob/master/src/main/java/de/komoot/photon/nominatim/NominatimConnector.java#L230) not only by geometry_sector but also by parent_place_id and then maintaining a one-item cache of the last parent address retrieved for a rank 30 object.

Just a rough estimate for the potential savings: there are 140 mio. rank 30 objects in a planet database but they have only 17 mio. unqiue parent_place_ids between them.

alfmarcua:
For the queries of the following type:
```sql
SELECT p.name, p.class, p.type, p.rank_address
FROM placex p, place_addressline pa
WHERE p.place_id = pa.address_place_id and pa.place_id = $1
and pa.cached_rank_address > 4 and pa.address_place_id != $2 and pa.isaddress
ORDER BY rank_address desc, fromarea desc, distance asc, rank_search desc
```
In order to speed up the process, another suggestion is to group them in batch or in a single query. This could even be  included in the query performed in readEntireDatabase method, see [NominatimConnector:228](https://github.com/komoot/photon/blob/2bbce026b58b44d33388d9bca0fe1747c32a9933/src/main/java/de/komoot/photon/nominatim/NominatimConnector.java#L228).

alfmarcua:
Another approach could relay in avoid the query for the term of the parent place ID itself, see [NominatimConnector:189](https://github.com/komoot/photon/blob/2bbce026b58b44d33388d9bca0fe1747c32a9933/src/main/java/de/komoot/photon/nominatim/NominatimConnector.java#L189). Maybe that data can be cached or can be get altoghether with the previous query. If those queries are avoided, it could save up about a 7% of the total query time.

lonvia:
Resolved by #547.

PR #547, opened 2021-03-16
Cache parent address for rank 29 and 30 (#529)

alfmarcua:
Parent address are cached and reused between ranks 29 and 30 objects with the same parent as indicated in #529

### DIFF
--- a/src/main/java/de/komoot/photon/nominatim/NominatimConnector.java
+++ b/src/main/java/de/komoot/photon/nominatim/NominatimConnector.java
@@ -162,6 +162,9 @@ public class NominatimConnector {
         return result.getDocsWithHousenumber();
     }
 
+    private long parentPlaceId = -1;
+    private List<AddressRow> parentTerms = null;
+
     List<AddressRow> getAddresses(PhotonDoc doc) {
         RowMapper<AddressRow> rowMapper = (rs, rowNum) -> new AddressRow(
                 dbutils.getMap(rs, "name"),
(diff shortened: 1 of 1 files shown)


alfmarcua:
I have made a few test with a Nominatim database with Portugal data. I have run `java -jar photon-0.3.4.jar -nominatim-import -host localhost -port 5432 -database nominatim_portugal -user nominatim -password XXXX -languages es,fr` for both versions.

The import of new version seems a bit faster (3 minutes 31 seconds) than the one performed with the master branch (4 minutes 54 seconds).

Also after logging all the queries made to the database, the logs of the newer version are smaller: 988 MB for the new version vs 1.8 GB for the master branch. Analyzing them gives 439585 queries for the new version vs 796212 for the master branch.

lonvia:
Looks good. I'm testing this on a planet, which will take a couple of days.

In the meantime, could you please rebase on master? I've fixed the CI issue there. So after rebasing the failing check should be gone.

alfmarcua:
Sure, rebase done!
We are dealing with speeding up the process and, from our results, we do not see a clear improvement of the overall process time despite of the number of queries has been significantly reduced.
Could you provide us with some insights about the obtained times and number of queries both before and after the pull request?

src/main/java/de/komoot/photon/nominatim/NominatimConnector.java
```diff
@@ -227,7 +244,7 @@ public void readEntireDatabase(String... countryCodes) {
 
         template.query(SELECT_COLS_PLACEX + " FROM placex " +
                 " WHERE linked_place_id IS NULL AND centroid IS NOT NULL " + andCountryCodeStr +
-                " ORDER BY geometry_sector; ", rs -> {
+                " ORDER BY parent_place_id, geometry_sector; ", rs -> {
```
lonvia: The order should be better ' geometry_sector, parent_place_id'. This ensures that subsequent places are geographically close. This makes a bit of a difference for postgresql because data can be better cached. More surprisingly, it does make a huge difference for ElasticSearch. If the data is not sorted geographically during insert then the database is 20% larger.

lonvia (commented):
For dumping the planet I get 8h 44min on master, 8h 04min with this PR and 7h 27min with this PR and reversed ORDER BY. So that is about 15% improvement. I don't have numbers for the number of queries.

src/main/java/de/komoot/photon/nominatim/NominatimConnector.java
alfmarcua: Thank you for the explanation. We actually did test both ways for Portugal import. Nevertheless, we did not find a huge impact on performance nor in the number of queries made, since the process does not last long enough to reflect that.

lonvia:
I've run another test on the planet after fixing the issue of the missing housenumbers (#558). Now the photon database is back at the expected 140GB. Master needs 21h 18min and this PR 12h 25min. That comes very close to the theoretical maximum of what the PR could achieve. Nice.

PR #553, opened 2021-03-24
Add lombok installation instructions for a develop environment

alfmarcua:
In a develop enviroment with an IDE like eclipse, lombok is needed.


lonvia:
I've decided to get rid of lombok instead.

PR #556, opened 2021-03-31
Bug/Reproduction case: NullPointerException during Photon update after updating Nominatim

leonardehrenfried:
This isn't a PR pre se but rather a reproduction case for my problem: I've noticed that when I tried to continuously update my Photon instance with updates from Nominatim it would fail because an NPE is thrown in `NominatimConnector`.

```
2021-03-31 11:01:26,292 [main] INFO  de.komoot.photon.nominatim.NominatimUpdater - Starting rank 1
2021-03-31 11:01:26,730 [main] INFO  de.komoot.photon.nominatim.NominatimUpdater - Starting rank 2
2021-03-31 11:01:26,738 [main] INFO  de.komoot.photon.nominatim.NominatimUpdater - Starting rank 3
2021-03-31 11:01:26,745 [main] INFO  de.komoot.photon.nominatim.NominatimUpdater - Starting rank 4
2021-03-31 11:01:26,751 [main] INFO  de.komoot.photon.nominatim.NominatimUpdater - Starting rank 5
2021-03-31 11:01:26,759 [main] INFO  de.komoot.photon.nominatim.NominatimUpdater - Starting rank 6
2021-03-31 11:01:26,766 [main] INFO  de.komoot.photon.nominatim.NominatimUpdater - Starting rank 7
2021-03-31 11:01:26,774 [main] INFO  de.komoot.photon.nominatim.NominatimUpdater - Starting rank 8
2021-03-31 11:01:26,780 [main] INFO  de.komoot.photon.nominatim.NominatimUpdater - Starting rank 9
2021-03-31 11:01:26,785 [main] INFO  de.komoot.photon.nominatim.NominatimUpdater - Starting rank 10
2021-03-31 11:01:26,794 [main] INFO  de.komoot.photon.nominatim.NominatimUpdater - Starting rank 11
2021-03-31 11:01:26,801 [main] INFO  de.komoot.photon.nominatim.NominatimUpdater - Starting rank 12
2021-03-31 11:01:26,809 [main] INFO  de.komoot.photon.nominatim.NominatimUpdater - Starting rank 13
2021-03-31 11:01:26,815 [main] INFO  de.komoot.photon.nominatim.NominatimUpdater - Starting rank 14
2021-03-31 11:01:26,824 [main] INFO  de.komoot.photon.nominatim.NominatimUpdater - Starting rank 15
2021-03-31 11:01:26,830 [main] INFO  de.komoot.photon.nominatim.NominatimUpdater - Starting rank 16
[...]

PR #558, opened 2021-04-02
Do not drop unnamed housenumber objects

lonvia:
By moving the usefulness check from the NominatimDoc to the PhotonDoc, objects with only a housenumber but no name got suddenly filtered. This reinstates the previous code, which tested for usefulness before adding the document.

Also fixes #556.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,4 @@
-name: "Continuous Integration"
+name: CI
 
 on:
   push:
@@ -35,6 +35,7 @@ jobs:
 
 
   import:
+    name: Import OSM data from Nominatim into Photon
     runs-on: ubuntu-20.04
 
     steps:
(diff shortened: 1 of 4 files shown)


leonardehrenfried:
Thanks for looking into this!

By using the cached version of the Monaco extract aren't you inherently relying on the fact that that is an unspecified number of days old rather than making sure that it is?

lonvia:
It doesn't matter if we grab the newest version of the Monaco abstract. When nothing is specified, then the updates use the OSM planet minutely updates as a default. Given that the Monaco extract already is a couple of hours outdated when it is published, it is guaranteed that there are always some updates to fetch, even with the latest one.

The caching is really only there to not annoy the friendly folks at Geofabrik too much when there is a lot of PR activity. ;)

leonardehrenfried:
Ah, thanks for the explanation! I did wonder how replication works when the URL is not explicitly configured. I thought it is read from the pbf but your explanation makes more sense as it explains why the nominatim update is so slow.

If you'd like me to configure the URL for the test let me know.

lonvia:
The minutelies are a better choice here, because a small extract like Monaco might not see any changes in a whole day. I've reduced the size of the update (the NOMINATIM_REPLICATION_MAX_DIFF variable) to make it reasonably fast.

Issue #574, opened 2021-04-29
Export additional information about OSM objects in Photon API


lonvia:
Nominatim has the capability to save any OSM tag as extra information with a place object and [to return this via the API](https://nominatim.org/release-docs/develop/api/Search/#output-details). Photon currently ignores this information completely even though it might be interesting for special purpose installations (e.g. a search engine for restaurants also wants to return the `cuisine` information if available.)

Returning arbitrary tags is a bit difficult with ElasticSearch but we could add a fixed set of selected tags. These would not be searchable but only be returned as extra information, just like Nominatim does. I'd suggest an additional property `extra`, which contains the tags that are available. If there is not extra infromation for a place, the property would be omitted completely, e.g

```
{
  "features": [
    {
      "geometry": {   "coordinates": [  11.5500582,  48.1778972], "type": "Point"  },
      "type": "Feature",
      "properties": {
        "osm_id": 4068438,
        "country": "Germany",
        "city": "Munich",
        "countrycode": "DE",
        "type": "house",
        "osm_type": "N",
        "osm_key": "amenity",
        "osm_value": "restaurant",
        "name": "Alphut",
        "extra": {
           "cuisine" : "vegan"
        }
      }
    },
      "geometry": {   "coordinates": [  11.5500582,  48.1778972], "type": "Point"  },
      "type": "Feature",
      "properties": {
        "osm_id": 291856,
        "country": "Germany",
        "name": "Munich",
        "osm_type": "place",
        "osm_value": "city"
      }
     }
  ]
}
```
Implementation steps:
* [ ] Provide an additional parameter `-extratags` for the import that selects the tags to add to the Photon database.
* [ ] Dynamically change the mapping to add fields for the additional tags.
* [ ] Add extra information to the JSON output.
* [ ] Add tests for the new function.

CC @hbruch

lonvia:
Implemented in #576.

PR #576, opened 2021-05-04
Add new import option '-extra-tags'

lonvia:
Implements #574.

The '-extra-tags' command line option takes a comma-separated list of OSM keys to be copied over from the Nominatim database. 

When running updates, the user currently has to take care to include the '-extra-tags' again in the command line. We could probably scan the ES mapping instead to retrieve the configured list but I'll leave that for another PR.


lonvia:
This is the wrong way to go about adding the mapping for the 'extra' field. We should be using [dynamic templates](https://www.elastic.co/guide/en/elasticsearch/reference/5.6/dynamic-templates.html) to simply add the required mappings on the fly. With such a template it is possible to ensure that the field is not indexed.

leonardehrenfried:
For this to work do you need the Nomatim `extratags` import style? 
And do you need a specific version of Nominatim?

lonvia:
Well, you need to make sure that the tags you are interested in have been previously imported into Nominatim. Using the `extratags` import style is one option. That style will ensure that most of the tags of an OSM object can be found in the extratags column. You can also start with one of the other styles and edit it to copy only the tags you need. Something like that:

```
{
    "keys" : ["cuisine", "takeaway", "website"],
    "values" : {
        "" : "extra"
    }
}
```

The new options works with all Nominatim versions that are usable with Photon.

FaFre:
@lonvia this is a great feature! Helps me a lot, thank you very much :)

Issue #310, opened 2018-01-30
Nominatim postcodes missing in Photon


hbruch:
The following Photon UK postcode searches return better matches in Nominatim.

* HP22 5NF - Upper Icknield Way, Aston Clinton, Aylesbury HP22 5NF, UK : https://photon.komoot.de/api?q=HP22%205NF vs https://nominatim.openstreetmap.org/search.php?q=HP22+5NF
* SN11 8AW - Oxford Rd, Calne SN11 8AW, UK : https://photon.komoot.de/api?q=SN11%208AW vs https://nominatim.openstreetmap.org/search.php?q=SN11%208AW
* SN2 1AE - Cricklade Rd, Swindon SN2 1AE, UK : https://photon.komoot.de/api?q=SN2%201AE vs https://nominatim.openstreetmap.org/search.php?q=SN2%201AE
* OX10 7LY - Dorchester, Wallingford OX10 7LY, UK : https://photon.komoot.de/api?q=OX10%207LY vs https://nominatim.openstreetmap.org/search.php?q=OX10%207LY
* BS15 4TA - Bristol BS15 4TA, UK : https://photon.komoot.de/api?q=BS15%204TA vs https://nominatim.openstreetmap.org/search.php?q=BS15%204TA

When I view the Nominatim result's html source, I see the search finds a place:postcode, e.g.

```
...
{
        "osm_type": "P",
        "osm_id": "82631",
        "class": "place",
        "type": "postcode",
        "admin_level": null,
        "rank_search": "25",
        "rank_address": "5",
        "place_id": "181154732",
        "parent_place_id": "179866619",
        "housenumber": null,
        "country_code": "gb",
        "langaddress": "South Gloucestershire, BS15 4TA, Vereinigtes K\u00f6nigreich",
        "placename": "BS15 4TA",
        "ref": "BS15 4TA",
        "lon": "-2.489309238268",
        "lat": "51.475183842419",
        "importance": 0.23125,
        "addressimportance": "8.83048074613031",
        "extra_place": null,
        "addresstype": "place",
        "aBoundingBox": [
            "51.475133842419",
            "51.475233842419",
            "-2.489359238268",
            "-2.489259238268"
        ],
        "label": "Postcode",
        "name": "South Gloucestershire, BS15 4TA, Vereinigtes K\u00f6nigreich",
        "foundorder": -0.85464807461303
    }
...
```

[...]

PR #580, opened 2021-05-17
Rework location bias and query builder in general

lonvia:
This is work in progress to improve the location bias for searches. As the unusually high boost for name queries was interfering with the location scoring, the query part itself needed to be reworked as well.

In detail:
* Location bias is switched to an exponential decay function with the given distance in kilometers as the 0.8 weight limit. To ensure that well-known places show up in the results nonetheless, use a second weight function based on importance and use the max of the two functions.
* Introduce a new parameter ' zoom'  for the location bias, which describes the radius for which the bias is effective as a function of the map zoom level. The precise relation is 0.25km*2^(18 - zoom). The `location_bias_scale` now describes how much weight is given to important places. Sensible values are between 0.0 and 1.0.
* Weight function for importance switched to a simple linear decay function. This essentially means using importance as a direct weight factor with the twist that it must not become completely zero.
* Reintroduce a subquery against the name.ngram index. This ensures similar results as the previous heavy boost on name.raw.
* Add a subquery against collector.raw which gets a strong boost when the house number matches. Records with housenumbers usually have no name (the street not being in the name indexes), so that they get a huge malus over name matches. The boost offsets this malus.

The query structure is still not ideal but any further improvements require a change to the index mapping. I would like to defer that until after the next release.

With the modified weight functions we might even be able to get rid of painless. That would be a huge relief.

There is a global test instance (using last weeks planet dump) with the changes running at https://pagan.lonvia.de/ Feedback would be very much appreciated.

@hbruch @leonardehrenfried


lonvia:
Here are a couple of queries that still fail:

* https://pagan.lonvia.de/api?debug=1&q=alta (TF messes with the result)
* https://pagan.lonvia.de/api?q=Wien%20Lohwaggase%2010 (skipping a word ranks higher than typos)

lonvia:
The latest commit fixes an issue where alternative names (`alt_name`, `old_name`, etc.) were not searchable anymore.

kenseii:
@lonvia I think with the rework of the query builder searching with post code is no longer working. For example the results of `/api?q=289-1606` on the #585 PR(which is based on this PR) returns data totally different to the ones of photon.komoot.io which are quite accurate. Not sure if its a bug or if the support of post code search are going to be dropped.

lonvia:
Postcode search has never really worked because we don't import the artificial postcode from Nominatim, see #310. Current master gives you some random result that would happen to have the postcode (a river in your example) but not really a postcode result. So I don't really see that as a regression.

lonvia:
I've merged this now and deployed it on https://photon.komoot.io. We'll see if there is feedback from the wider user community.

PR #581, opened 2021-05-25
Experimental: simple synonym list

lonvia:
This adds a new command-line parameter `-synonym-file` which can be used to experiment with applying synonyms at query time. The synonym list is loaded when Photon is started. It is not necessary to reindex the existing database and will work against our precomputed database dumps.

The synonym file must be a simple text file. Each line contains a comma-separated list of mutual synonyms. Only single word
synonyms are allowed. To that end, the loader will reject any synonym file that contains spaces.

This PR is unlikely to make it into master as is. I just thought I put it out in public in case somebody else wants to experiment with using synonyms.

@hbruch

### DIFF
--- a/src/main/java/de/komoot/photon/elasticsearch/IndexSettings.java
+++ b/src/main/java/de/komoot/photon/elasticsearch/IndexSettings.java
@@ -72,7 +72,7 @@ public class IndexSettings {
             if (line.indexOf(' ') >= 0) {
                 throw new RuntimeException("Synonym list must not contain any spaces or multi word terms.");
             }
-            synonyms.put(line);
+            synonyms.put(line.toLowerCase());
         }
 
         JSONObject filters = (JSONObject) settings.optQuery("/analysis/filter");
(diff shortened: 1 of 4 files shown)


src/main/java/de/komoot/photon/elasticsearch/IndexSettings.java
```diff
@@ -72,7 +72,7 @@ public IndexSettings setSynonyms(String synonymFile) throws IOException {
             if (line.indexOf(' ') >= 0) {
                 throw new RuntimeException("Synonym list must not contain any spaces or multi word terms.");
             }
-            synonyms.put(line);
+            synonyms.put(line.toLowerCase());
```
otbutz, line 75: `line.toLowerCase(Locale.ROOT)` ?

lonvia: It should be "whatever ES uses", to be precise.

lonvia:
Now included in #585.

Issue #557, opened 2021-04-02
Add special phrase searching like nominatim


JinIgarashi:
I just found [this article](https://nominatim.org/release-docs/latest/admin/Import/#tuning-the-database) in nominatim website.

> If you want to be able to search for places by their type through special key phrases you also need to enable these key phrases like this:
```
./utils/specialphrases.php --wiki-import > specialphrases.sql
psql -d nominatim -f specialphrases.sql
```
> Note that this command downloads the phrases from the wiki link above. You need internet access for the step.

I think it might be better if we can search POI by key from [Nominatim/Special Phrase](https://wiki.openstreetmap.org/wiki/Nominatim/Special_Phrases). For instance, search `bank` type only at certain city.

I am not very sure whether it is possible technically. What do you think?

lonvia:
It's technically possible but we are a long way off from a practical implementation.

JinIgarashi:
> It's technically possible but we are a long way off from a practical implementation.

Thanks for the comments. This is just my humble suggestion. It would be great if photon can consider particular phrases to prioritize searching.

kenseii:
@lonvia we are thinking about implementing something similar to what is being discussed in this issue.

Basically when searching or ordering the results, we would like some `osm_values` e.g: ['aerodrome', 'station','stop'] to be boosted, scored higher so that they show up at the top of the results compared to e.g: ['hotel','sauna'] even if the osm importance of the later would be higher.

Any idea or recommendation on how to implement this or where to look?

Is it better to perform this when querying, scoring or returning the data?

Is it something worth submitting a PR to the photon repo?

Thank you

lonvia:
@kenseii It sounds like you are looking rather for a static boost by OSM type, i.e. the boost would be independent from the actual query. This issue is more about searching by keywords. 'tokyo station' would boost train stations, 'tokyo hotel' would boost hotels.

kenseii:
@lonvia In order to make the search results dynamically boosted depending on the `osm_value`'s type of the query, i am thinking of using synonyms. 

e.g: if i search for Tokyo station, i would like to boost the results that have "station" as the `osm_value`.

the reason why we think synonyms are important is because the search query might hold a useful key that is not an osm_value

e.g searching with Narita airport: 
Narita airport -> airport -> aerodrome, so we boost the result with aerodrome

e.g searching with Narita hotel:
Narita hotel -> hotel -> hotel, so we boost the result with hotel

[...]

PR #585, opened 2021-06-11
Synonym list + classification terms

### LINKED ISSUE #318: Improve POI search
When we search for a railway station we have the problem in German that often the most important station is Hauptbahnhof but the match to "Bahnhof" is much worse although it should be much better. 

Also there is a problem for different languages like "<city name> railway station" ... so we should think to index common POI names additionally to a POI for the different languages (airport, railway, post office, ...)

lonvia:
This PR extends #581 and adds classification terms. The classification terms enable searches where the descriptive term of the place is given in addition to the name ('London airport', 'Louvre museum', ...).  Like synonyms, the classification terms are restricted to single word terms. This is a restriction in the synonym module of ES5. See `docs/synonyms.md` for more information on the new format of the synonym file.

Classification terms work by adding a cryptic term that is derived from the OSM key and value and usually does not interfere with any other search terms. The cryptic term is added to the collector and then we define synonyms from the actual classification terms to the cryptic term. This means that we can stay with a single query and have ES decide if a term should be interpreted as a classification term or as part of a name. It also means that classification terms can be changed on an existing database, given that they are just query time synonyms.

As classification terms need to be newly indexed, a mapping schema change and reimport is required and the database version has increased.

Partially implements #557. Classification terms cannot be used for searching by type ('museum in Paris', 'airport near London').

Fixes #318.

### DIFF
--- a/docs/synonyms.md
+++ b/docs/synonyms.md
@@ -1,0 +1,83 @@
+# Using Synonyms and Classification Terms
+
+Photon has built-in support for using custom query-time synonyms and
+special phrases for searching a place by its type. This document explains
+how to configure this feature.
+
+## Configuration
+
+Synonyms and classification terms are configured with a JSON file which can
+be added to a Photon server instance using the command line parameter
+`-synonym-file`. Synonyms are a run-time feature. Handing in a synonym list
+at import time has no effect. The list of synonyms in use can simply be
+changed by restarting the Photon server with a different synonym list (or
+not at all, if you want to completely disable the feature again).
+
+Here is a simple example of a synonym configuration file:
+
+```
+{
+  "search_synonyms": [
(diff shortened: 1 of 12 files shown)


kenseii:
> See `docs/synonyms.md` for more information on the new format of the synonym file

Sorry I cant find this file, do you mind sharing its path?
It seems like the normal .txt synonym files are no longer supported.

lonvia:
Forgot to push the documentation file. Fixed now.

lonvia:
The last commit fixes a small error that prevented results that match the classificiation terms from being boosted. 

@leonardehrenfried you might want to update your branches again to get the last commit 0ac89a5. This change only affects searches, no reimport of the index necessary.

Issue #592, opened 2021-07-05
Filter extra tags with osm_tag


FaFre:
Since #576 we can include additional information in the results. This is a great feature, and saves from reloading of full elements a lot of times :+1: 
Would the data structure also allow to also include the extra tags in the already existing search-filter for `osm_tag`?

E.g.: `api?q=berlin&osm_tag=cuisine`

This would be a great addition :)

lonvia:
The extra data is currently not indexed and thus not searchable. In any case, if we wanted to search for it, this would require some special infrastructure in the query builder. Not impossible but not very high on the TODO list at the moment.

FaFre:
Has there been a PR recently, where another property got indexed? So I can get an idea about the effort required.

henrik242:
Just dropping by to show my support for this issue. Filtering based on the `extra` properties is a necessity in our usecase.

lonvia:
The most straight-forward solution here would be to have a [dynamic mapping template](https://docs.opensearch.org/docs/latest/field-types/#dynamic-templates) for the extra field, so that we end up with one index per key. Based on that have an `extra` parameter that has the same syntax as the `osm_tag` parameter. Only the value filters are not implementable but that is no loss. #594 goes somewhat in that direction but is missing support for arbitrary keys and for querying key-value parameters.

That solution works probably okay as long as there is a restricted set of extra tags. It is unlikely to scale properly once we import all possible extra tags from Nominatim. That's why I was thinking to experiment with [doc values](https://docs.opensearch.org/docs/latest/field-types/mapping-parameters/doc-values/) instead. Basically the idea here is that the main result selection happens through the query and we can just do some lightweight result filtering by extra tags instead of doing a full index lookup. It is not yet clear to me if doc values can actually achieve that. And even if it works, the question is, what the difference is in terms of query speed and database size.

henrik242:
PR up for discussion: https://github.com/komoot/photon/pull/908

Feedback very welcome!

lonvia:
Filtering by extra tags will not be possible, they are display information only. Use [categories](https://github.com/komoot/photon/blob/master/docs/categories.md) when custom filtering is needed.

PR #594, opened 2021-07-07
Filtering of extra tags

FaFre:
Draft PR for filtering of extra tags, related to #592.

So far it only enables filtering on the existence of keys inside the extra-tag object. Extra tags has been introduced via #576.

Example: http://localhost:2322/api?q=monaco&extra_key=wikidata&extra_key=wikipedia
Will only return elements with `wikidata` or `wikipedia` tags in extra. Behavior is similar to `osm_key`.

ToDo:
- So far only `wikipedia` and `wikidata` gets indexed in elastic, because I don't know how to make it dynamic for all properties inside `extra`. If the mapping has to get done explicitly, all the extra tags can be found here: https://taginfo.openstreetmap.org/projects/nominatim#tags
- No tests has been written yet (I'm not a java developer and don't want to deal with the tests for now...)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,10 @@ nbactions.xml
 .classpath
 .project
 .settings/
+.factorypath
+
+##############################
+## Visual Studio Code
+##############################
+.vscode/
+.code-workspace
--- a/es/mappings.json
+++ b/es/mappings.json
@@ -46,6 +46,16 @@
 					}
 				}
 			},
+			"extra": {
+				"properties": {
+					"wikidata": {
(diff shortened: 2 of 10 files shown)


phamtai97:
Any update PR? We need to your feature to query by extra tags because we has add some tags to extra-tag and want to search by it?

lvidaller:
We need to filter using extra tag also. It could be very useful !

FaFre:
There will be no update from my side anymore unfortunately, maybe someone wants to pick the work up again.

Issue #609, opened 2021-10-25
Missing elements in address line while importing


rocainunwired:
`place_addressline` table in Nominatim DB does not have the entire address line for a few addresses. This causes missing address elements while importing data into photon. (Unsure if this is a bug on Nominatim's side)

Could we instead use `get_addressdata` DB function used by Nominatim DB to fetch the address line?

Ex. 17.42067, 78.40256
https://photon.komoot.io/reverse?lat=17.42067&lon=78.40256
vs 
https://nominatim.openstreetmap.org/reverse.php?lat=17.42067&lon=78.40256&zoom=18&format=json

You can notice the missing address elements like "Hyderabad" (city), "Jubilee Hills" (suburb)

PR #617, opened 2021-12-09
Take geography into account for address computation

lonvia:
Use the more complex way of computing the address from Nominatim for computing the rank 30 address. That makes sure that address parts are not blindly taken from the parent but geographic relations are taken into account. When streets go over boundaries, the addresses along that street will be corrected as necessary according to the administrative entity that contains them. See also Nominatim PR https://github.com/osm-search/Nominatim/pull/2082

Requires advanced SQL which is not supported by H2. Thus use the previous simpler query for tests instead.

Sadly, this PR is incompatible with #547. If it gets merged, then import time will again increase by around 50%. So it  would be useful if others tested this a bit to figure out if it is worth the performance loss. I haven't made up my mind yet if I want to merge it or not.

See also discussion in #609.

### DIFF
--- a/src/main/java/de/komoot/photon/nominatim/DBDataAdapter.java
+++ b/src/main/java/de/komoot/photon/nominatim/DBDataAdapter.java
@@ -21,4 +21,9 @@ public interface DBDataAdapter {
      */
     @Nullable
     Geometry extractGeometry(ResultSet rs, String columnName) throws SQLException;
+
+    /**
+     * Return the SQL for computing the address terms of a place.
+     */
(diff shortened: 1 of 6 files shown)


rocainunwired:
I pulled this branch, compiled with `-DskipTests=true` as the tests were failing.

Loaded the Nominatim DB with data only from Hyderbad, India.

Total no of records imported by photon: 18983
Import time with master branch: 40 secs (475 records/sec)
Import time with this branch: 43 secs (442 records/sec)

The address hierarchy issue appears to be fixed with this branch on testing a few data points.

lonvia:
This isn't really working as intended. Trying a different approach.

PR #630, opened 2022-01-12
Map equality check should use equal

hendrikmoree:
It seem this compare is not always right. `==` compares the same object, but sometimes it a different object with the same contents.

So it's beter to use `.equals` on the map.

All tests seems to work, and I've added some extra tests to prove it.

### DIFF
--- a/src/main/java/de/komoot/photon/PhotonDoc.java
+++ b/src/main/java/de/komoot/photon/PhotonDoc.java
@@ -216,7 +216,8 @@ public class PhotonDoc {
      * @return True, if the address was inserted.
      */
     public boolean setAddressPartIfNew(AddressType addressType, Map<String, String> names) {
-        return addressParts.computeIfAbsent(addressType, k -> names) == names;
+        Map<String, String> value = addressParts.computeIfAbsent(addressType, k -> names);
+        return value != null && value.equals(names);
     }
 
     public void setCountry(Map<String, String> names) {
--- a/src/test/java/de/komoot/photon/PhotonDocTest.java
+++ b/src/test/java/de/komoot/photon/PhotonDocTest.java
@@ -40,6 +40,51 @@ public class PhotonDocTest {
         Assert.assertEquals("DE", doc.getCountryCode().getAlpha2());
     }
 
+    @Test
+    public void testSetNewAddressPart() {
+        PhotonDoc doc = simplePhotonDoc();
+
+        HashMap<String, String> streetNames = new HashMap<>();
+        streetNames.put("name", "parent place street");
+        Assert.assertTrue(doc.setAddressPartIfNew(AddressType.STREET, streetNames));
+    }
+
+    @Test
+    public void testSetOtherAddressPart() {
+        PhotonDoc doc = simplePhotonDoc();
+
+        HashMap<String, String> streetNames = new HashMap<>();
+        streetNames.put("name", "parent place street");
+        Assert.assertTrue(doc.setAddressPartIfNew(AddressType.STREET, streetNames));
+
+        HashMap<String, String> streetNames2 = new HashMap<>();
+        streetNames2.put("name", "parent place street 2");
+        Assert.assertFalse(doc.setAddressPartIfNew(AddressType.STREET, streetNames2));
+    }
+
+    @Test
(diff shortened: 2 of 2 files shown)


lonvia:
The object comparison is intentional here. The function returns true if and only if the given name hash was inserted into `addressParts`. A different object with the same contents still wouldn't be inserted.

hendrikmoree:
Mm okay. But there's something going on for example for this node: https://www.openstreetmap.org/node/45247254

In this case 'Utrecht' is inserted into the context, even if 'Utrecht' is already available in the state. This gives some unexpected result at the end. I don't know exactly why a different object with the same content is passed for this node, maybe it's a different issue?

This is because a different object, with the same context is passed here: https://github.com/komoot/photon/blob/ce84ef47ed85298bd9986d005523a685b7636cc6/src/main/java/de/komoot/photon/nominatim/NominatimConnector.java#L279

This doesn't happen for a lot of other objects inside the same state. Is that also intentional?

lonvia:
This node looks all correct to me. [Photon](https://photon.komoot.io/api/?q=Voskuilen) reports city=Woudenberg + state=Utrecht which is exactly what [Nominatim](https://nominatim.openstreetmap.org/ui/details.html?osmtype=N&osmid=45247254) has. There are no addr: tags on that node, so the address parts are 100% computed.

hendrikmoree:
Yes sure that's true. But for some reason Utrecht is also added to the context field, which is not necessary because it's already in state.

lonvia:
Ah, I see. We should indeed figure out why there are multiple Utrechts returned by `getAddresses()`. You wouldn't have a some simple instructions to reproduce the issue?

hendrikmoree:
No not really, I discovered it because we rank state very low and don't add it to other fields.

I tried it by only updating this node (with a custom Nominatim import as wel 🙈 ) and logging the contents.

lonvia:
Might be an issue in Nominatim. What does the raw SQL query for addresses return:

```
SELECT osm_type, osm_id, class, type, rank_address, name->'name' FROM placex p, place_addressline pa WHERE p.place_id = pa.address_place_id and pa.place_id = ??? and pa.cached_rank_address > 4 and pa.address_place_id != ??? and pa.isaddress ORDER BY rank_address desc, fromarea desc, distance asc, rank_search desc;
```
where `???` needs to be replaced with the right place_id (`select place_id from placex where osm_type = 'N' and osm_id = 45247254`).

hendrikmoree:
Seems okay to me:
```
 osm_type | osm_id |  class   |      type      | rank_address |  ?column?
----------+--------+----------+----------------+--------------+------------
 R        | 419036 | boundary | administrative |           16 | Woudenberg
 R        |  47667 | boundary | administrative |            8 | Utrecht
 R        |  47796 | boundary | administrative |            6 | Nederland
(3 rows)
```

hendrikmoree:
It seems to happen because completePlace is called twice on updating. The second time you've another object for the same address. 

It happens on these lines:
https://github.com/komoot/photon/blob/ce84ef47ed85298bd9986d005523a685b7636cc6/src/main/java/de/komoot/photon/nominatim/NominatimConnector.java#L83

https://github.com/komoot/photon/blob/ce84ef47ed85298bd9986d005523a685b7636cc6/src/main/java/de/komoot/photon/nominatim/NominatimConnector.java#L148

Why is this called twice? It that on purpose?

lonvia:
That looks like a bug. This piece of code is only called for updates, so we won't see the problem with normal extracts.

hendrikmoree:
Thanks! I think this PR can be closed. And so it is safe to remove this line? https://github.com/komoot/photon/blob/ce84ef47ed85298bd9986d005523a685b7636cc6/src/main/java/de/komoot/photon/nominatim/NominatimConnector.java#L148
And I think also this line:
https://github.com/komoot/photon/blob/ce84ef47ed85298bd9986d005523a685b7636cc6/src/main/java/de/komoot/photon/nominatim/NominatimConnector.java#L157

lonvia:
Yes, I think so.

I've transferred this discussion in to a new issue as this definitely needs fixing. I'd just would like to see some tests for it. Our test coverage for updates is really bad. Closing here.

PR #650, opened 2022-03-01
Ensure ImportThread is always terminated.

codepainters:
This pull request solves the problem I observe with v0.3.5: if `-nominatim-import` is executed when the database is not reachable, the process never ever exits. This is due to `ImportThead.finish()` never being called, and thus `ImportThread` never terminating. 

Note: this issue is no longer observable with latest code. After https://github.com/komoot/photon/commit/45b3ec75356bf795ddb63b91a734e1eb49f74801 commit `NominatimConstructor` reaches to the database (and fails) before `ImportThread` is created, effectively masking the problem. 

However it can still fail, should any of the queries in `readEntireDatabase()` raise any exception.


lonvia:
Makes perfect sense. Thank you!

PR #652, opened 2022-03-04
Get rid of Guava dependency

lonvia:
Most of the constructs that Photon uses from guava have a equivalent built-ins in  Java 8+. So no need to keep this dependency around.

The handling of sets in the tests for PhotonRequestFactory becomes a bit awkward. That's better fixed with improving the interface of PhotonRequest, though.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -139,11 +139,6 @@
            </exclusions>
         </dependency>
         <dependency>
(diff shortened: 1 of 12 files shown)


simonpoole (approved):
Looks good to me.

PR #657, opened 2022-03-08
Fix geojson geometry type in GeocodeJsonFormatter

### LINKED ISSUE #656: geometries are of type feature instead of Point
With the latest addition of the GeocodeJsonFormatter file, the geojson feature collection contains features of type 'feature' instead of 'Point' as previously.

Here's an example of the geojson produced by the api:
```
{
    "features": [
      {
        "geometry": {
          "coordinates": [
            -73.53425581561339,
            45.51415475
          ],
          "type": "feature"
        },
        "type": "feature",
        "properties": {
          "osm_type": "R"
          <more properties>
        }
      }
    ],
    "type": "FeatureCollection"
}
```

tahini:
Fixes #656

Also use the constants instead of hard-coded strings in the
GeocodeJsonFormatter class and add unit tests for the geojson formatter.

### DIFF
--- a/src/main/java/de/komoot/photon/searcher/GeocodeJsonFormatter.java
+++ b/src/main/java/de/komoot/photon/searcher/GeocodeJsonFormatter.java
@@ -30,10 +30,10 @@ public class GeocodeJsonFormatter implements ResultFormatter {
             final double[] coordinates = result.getCoordinates();
 
             features.put(new JSONObject()
-                        .put("type", "feature")
+                        .put("type", "Feature")
                         .put("properties", getResultProperties(result))
                         .put("geometry", new JSONObject()
-                                .put("type", "feature")
+                                .put("type", "Point")
                                 .put("coordinates", coordinates)));
         }
 
--- a/src/test/java/de/komoot/photon/searcher/GeocodeJsonFormatterTest.java
+++ b/src/test/java/de/komoot/photon/searcher/GeocodeJsonFormatterTest.java
@@ -1,0 +1,45 @@
+package de.komoot.photon.searcher;
+
+import de.komoot.photon.Constants;
+import org.json.JSONObject;
+import org.json.JSONArray;
+import org.junit.jupiter.api.Test;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+public class GeocodeJsonFormatterTest {
+
+    @Test
+    public void testConvertToGeojson() {
+        GeocodeJsonFormatter formatter = new GeocodeJsonFormatter(false, "en");
+        List<PhotonResult> allResults = new ArrayList<>();
+        allResults.add(createDummyResult("99999", "Park Foo", "leisure", "park"));
+        allResults.add(createDummyResult("88888", "Bar Park", "leisure", "park"));
+
+        String geojsonString = formatter.convert(allResults, null);
+        JSONObject jsonObj = new JSONObject(geojsonString);
+        assertEquals("FeatureCollection", jsonObj.getString("type"));
+        JSONArray features = jsonObj.getJSONArray("features");
+        assertEquals(2, features.length());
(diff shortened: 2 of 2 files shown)


src/main/java/de/komoot/photon/searcher/GeocodeJsonFormatter.java
```diff
@@ -30,15 +30,15 @@ public String convert(List<PhotonResult> results, String debugInfo) {
...
+                        .put(Constants.TYPE, Constants.FEATURE)
+                        .put(Constants.PROPERTIES, getResultProperties(result))
+                        .put(Constants.GEOMETRY, new JSONObject()
+                                .put(Constants.TYPE, Constants.POINT)
+                                .put(Constants.COORDINATES, coordinates)));
         }
 
         final JSONObject out = new JSONObject();
-        out.put("type", "FeatureCollection")
+        out.put(Constants.TYPE, "FeatureCollection")
```
simonpoole: Surely "FeatureCollection" should be a constant too?

PS: personally I would move all GeoJSON related constants to their own class, but that's a matter of taste.

tahini: Indeed, I'll add this constants too

lonvia: Hmm, I had rather intentionally stopped using constants when refactoring the code. Is Java doing some special optimizations with constant strings that I'm not aware of? Otherwise I'd rather prefer to go for readability here and keep the literal strings.

simonpoole: No special processing. Using string literals (actually any literal) is just not considered good Java style, particulary when they are potentially used more than once, but YMMV. 

In this case the other argument is that the values are third party defined standard values and making them constants allows to document their provenance (IDEs will typically show the actual value on hover or similar).

All that said  I was just asking for consistency in my comment.

PS: before this gets totally sidetracked, the actual fix would seem to be correct.

lonvia: The fix is great. That there are tests is even better. I'll have to ponder about the rest.

tahini: Personally, I tend to always extract literals to constants, it avoids typos and if it ever needs to be changed, there's only one place to do so and tests still pass! Anyway, I saw those constants existed and were used in the file that this one replaced, so I just thought to replace them all here. Let me know if you prefer the direct literals and I'll update the PR.

lonvia: So here is the thing: the current global Constants.java file is just bad, no matter what the Java coding conventions say. The constants are used all over the place, partially with the same constant referring to completely different concepts. As they are, they just represent a glorified way of writing a string literal. That makes any argument for the constants mute. There is no reference to what they mean, there is no 'just changing them in one place'. This file definitely has to go away at some point and I won't accept any new additions or uses of the constant.

That doesn't mean that I object the use of constants in general. There are many good arguments for them. I just don't see much of a point in this particular case, as they will be used mostly exactly once (except in the tests, but writing them there again would rather be an additional safeguard against typos in the literal) but I don't mind them being used either. Just either move them in a separate class (`GeoJSONConstants` or something like that) or have them as part of the `GeocodeJsonFormatter` class itself.

otbutz: Any reason why we don't use something like mapbox geojson lib for this?

tahini: Ok, I'll just backtrack and fix the issue at hand without changing too much. It's my first contribution on this project after all! :D

lonvia:
Thank you @tahini. 

And don't let the lengthy discussion discourage you, the PR is really appreciated.

PR #667, opened 2022-04-05
Filter by object type

macolu:
As discussed in #666, here comes a first attempt of adding a new **object_type** filter.

Thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,6 +180,24 @@ Or, just by they key
 http://localhost:2322/api?q=berlin&osm_tag=tourism
 ```
 
+#### Filter results by object type
+
+List of available types:
+
+- house
+- street
+- locality
+- district
+- city
+- county
+- state
+- country
+
+```
+http://localhost:2322/api?q=berlin&object_type=city&object_type=locality
+```
+
+
 #### Results as GeoJSON
 ```json
 {
--- a/es/mappings.json
+++ b/es/mappings.json
@@ -194,8 +194,8 @@
 				"index": true
 			},
 			"object_type": {
-				"type": "text",
-				"index": false
+				"type": "keyword",
+				"index": true
 			},
 			"postcode": {
 				"type": "text",
--- a/src/main/java/de/komoot/photon/ReverseSearchRequestHandler.java
+++ b/src/main/java/de/komoot/photon/ReverseSearchRequestHandler.java
@@ -48,6 +48,11 @@ public class ReverseSearchRequestHandler extends RouteImpl {
             results = results.subList(0, photonRequest.getLimit());
         }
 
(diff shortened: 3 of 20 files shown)


lonvia:
Github actions fixed. Please rebase on master.

macolu:
Rebase done.
Doc updated.

I agree that we can probably find a better name :-)

lonvia:
I had some more off-line discussions and the consensus was that Pelias' `layer` is most appropriate.

@macolu Can I ask you to change the parameter name to `layer`. You can also change the name of the index in the database, if that makes things easier. After that, I'm happy to merge.

lonvia:
> Attribute in returned results. Since attribute had previously been renamed since last 0.3.5 release (type -> object_type), I assume there is no additional BC break.

Oh dear. The renaming was not intentional. That's an accident of the recent code refactoring. Good thing that you spotted it. We should stick with 'type' for the output. There may already be clients that rely on that name.

macolu:
In that case, maybe query string param should be named `type` as well?

It feels inconsistent to have a param named `layer` that filters on an attribute named `type` in output results... Two different names for the same notion.

macolu:
Here comes a few changes:

- reverted attribute name to `type` in response, to fix accidental BC break. Achieved by renaming field in Elastic mapping. `type` is quite vague, so maybe we should keep `object_type` there and have some mapping in Json Formatter... Your call :)  
- renamed new input query param to `type`, to be consistent with response.

I kept `objectType` in most of the code, again because `type` is quite vague, so it felt wrong to rename everything.

lonvia:
Sorry for staling here. I needed to ponder a bit on this.

So `type` in the response comes from the [geocodejson spec](https://github.com/geocoders/geocodejson-spec/tree/master/draft), which means we are stuck with it for better or worse. It explains the 'object_type' column name though. Obviously I also thought that 'type' is a bit vague. However, in general I agree with your changes including renaming the ES field.

That leaves the API parameter. I'd still like to go with `layer` here, for the simple reason that I had in mind that it is not exactly the same as type but can be a superset. In particular, I was thinking right now of a special layer=address that solves https://github.com/osm-search/Nominatim/issues/1469. Photon has exactly the same issue here layer=house doesn't guarantee that you get a housenumber. You might also get a named postbox or fountain or other curious POI objects.

macolu:
Ok for me. So here comes a new version:
* `type` in response and ES documents
* `layer` in query param and most of the code

lonvia:
Merged. Thanks you for your patience.

If all goes well the change should already be included in the next [experimental dumps](https://download1.graphhopper.com/public/experimental/). That will be a good occasion to officially announce them.

Issue #620, opened 2021-12-10
RCE 0-day exploit in log4j


lonvia:
Last night a [0-day exploit](https://www.lunasec.io/docs/blog/log4j-zero-day/) was published for log4j. This issue is to track impact for photon.

Current situation
----------------------

We currently use log4j in version 2.8 and cannot update to a later version due to ES 5.6 not being compatible with those. This means that photon is affected and it is yet unclear if a patched version will be available.

A preliminary code analysis shows that the security risk for photon as a result of this exploit is low. Photon does not log any data received through requests via log4j itself. The only user-supplied data logged are filenames and data from the ES database in rare cases. However, photon does print exception messages and we cannot rule out that any of the libraries we use will print a message that is subject to the exploit.

karussell:
[On HN](https://news.ycombinator.com/item?id=29504964) someone suggested to use:

```
log4j.formatMsgNoLookups=true or log4j2.formatMsgNoLookups=true
```

which we could suggest to use. 

> cannot update to a later version due to ES 5.6 not being compatible with those

Did you try this out or is this in their documentation somewhere?

kannes:
https://github.com/apache/logging-log4j2/pull/608#issuecomment-990305306 suggests one can also remove a class from the jar as "Holzhammer" solution.

lonvia:
> Did you try this out or is this in their documentation somewhere?

https://github.com/komoot/photon/pull/477#issuecomment-654505140

I'll see if it works to publish a jar without the class file.

lonvia:
Removing the JndiLookup.class is now the officially recommended mitigation according to the [Apache Log4j security page](https://logging.apache.org/log4j/2.x/security.html). To remove the class from your photon jar run:

    zip -q -d photon-*.jar org/apache/logging/log4j/core/lookup/JndiLookup.class

Then restart your photon server.

lonvia:
All jars downloadable from the [Relase page](https://github.com/komoot/photon/releases) for versions 0.3.x now have the JndiLookup.class removed. #621 makes sure that master builds remove the offending class when creating the final jar.

I consider the issue thus closed for Photon. Please make sure to update your servers.

PR #669, opened 2022-04-15
dependency update

dobrawek:
I've started working with photon - I've updated all dependencies to the lastest and changed usage of com.vividsolutions.jts to org.locationtech.jts that is now a part of latest postgis

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -304,8 +304,8 @@
         </resources>
     </build>
     <properties>
-        <maven.compiler.source>11</maven.compiler.source>
-        <maven.compiler.target>11</maven.compiler.target>
+        <maven.compiler.source>1.8</maven.compiler.source>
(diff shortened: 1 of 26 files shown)


lonvia:
Unlikely to work because ES 5.6 relies on com.vividsolutions.jts

avshabavsha:
Is it possible to exclude the versions update that are uncertain of and approve the rest?

Specifically, I planned to create a PR to update log4j to 2.17.2 (same version as @dobrawek did), but noticed it's already part of this PR.

This is to resolve these two issues:
https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-44832
https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-45105

lonvia:
See #620 for the mentioned CVEs. They have been long since resolved.

PR #701, opened 2022-11-02
Add link to PhotonDialog

woheller69:


lonvia:
I'm sorry but we are not taking PRs for the 'Related Projects' section at this point. The section is there more for historic reasons than anything. If it was to be revived, we'd need to come up with inclusion criteria first.

PR #733, opened 2023-05-25
Download signing key

otbutz:
https://wiki.postgresql.org/wiki/Apt#Quickstart


lonvia:
This solves the key issue but looking at the code I think it's better to simply switch to Ubuntu 22 and use the stock postgresql. Commit coming...

PR #734, opened 2023-05-25
Switch datanucleus to HTTPS

otbutz:
https://maven.apache.org/docs/3.8.1/release-notes.html#how-to-fix-when-i-get-a-http-repository-blocked


otbutz:
Otherwise the build fails for me while setting up a new Github workflow:

image

otbutz:
Both rebased

PR #741, opened 2023-07-14
Use OpenStreetMap Tiles

lukaswelte:
The referenced komoot tiles will be shut down soon.
To keep a map on the website please switch to public OpenStreetMap tiles

This only changes the example config, likely the actually used config is private - so please switch it there as well :) 

I'm happy to answer any kind of questions.
Thank you

### DIFF
--- a/website/config.js.example
+++ b/website/config.js.example
@@ -1,4 +1,4 @@
 API_URL = 'http://localhost:5001/api/?';
-TILELAYER = 'https://tile.openstreetmap.org/{z}/{x}/{y}.png;
+TILELAYER = 'https://tile.openstreetmap.org/{z}/{x}/{y}.png';
 CENTER = [52.3879, 13.0582];
 MAXZOOM = 18;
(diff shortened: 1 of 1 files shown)


otbutz:
Please use `tile.openstreetmap.org/{z}/{x}/{y}.png`. see https://github.com/openstreetmap/operations/issues/737

lukaswelte:
@otbutz thanks for reaching out - updated the PR

website/config.js.example
```diff
@@ -1,4 +1,4 @@
 API_URL = 'http://localhost:5001/api/?';
-TILELAYER = 'https://{s}.tile.hosted.thunderforest.com/komoot-2/{z}/{x}/{y}{r}.png';
+TILELAYER = 'https://tile.openstreetmap.org/{z}/{x}/{y}.png;
```
lonvia: Minor typo here. You deleted a ' too many.

lonvia:
I've updated photon.komoot.io accordingly.

Issue #703, opened 2022-11-14
Reverse search could support osm tag filters


karussell:
Would be nice if [osm tag filters](https://github.com/komoot/photon#filter-results-by-tags-and-values) are supported for reverse search too: https://photon.komoot.io/reverse/?lat=46.99483&lon=9.075354&osm_tag=amenity:pharmacy

lonvia:
That wouldn't be a reverse search anymore unless you are looking for a function "get me where I'm but only return a result if I happen to be in a pharmacy'. I suspect you are looking for a function "get me all pharmacies close to this location"?

kw005:
No, I have the location ID and need the data (location name, lat, lng etc.) for this ID. This is actually very important, e.g. a user is selecting location from the website via autocomplete and I will pass the location ID, lat, lng, name etc. to my server:

...[13.088345,52.6755087,13.7611609,52.3382448],"country":"Germany","osm_key":"place","countrycode":"DE","osm_value":"city","name":"Berlin","type":"city....

but those data (lng, lat, name) can me modified via data dumper or simple Browser Plugin in this https request. So, a user could change the location name "berlin" to "see my ad here" or f*ck that website" and I would enter those data into the database. That's a security issue and a vulnerable API. 

Better: I fetch the location ID only and get all other data (name, lat, lng etc.) from a second API call, made internally by the server. A curl request to an external site has also other issues (performance, no independency, DNS Firewall, API/layout change, too many curl calls for a big website etc. - just too many variables).

Is this somehow possible or could this be implemented to improve security?

mtmail:
@kw005 That has nothing to do with the issue "Reverse search could support osm tag filters". Open a new issue, or better discussion if you're reporting a different use case.

kw005:
@mtmail : Yeah, I saw that. Just deleted it ;)

lonvia:
In theory we have the `query_string_filter` parameter for reverse which got sneaked in in #254.  It should be able to do arbitrary filtering including the one you want. However, this was never documented and I very much prefer to let it die quietly and implement the tag filters from search instead.

I also wasn't aware that reverse already has a `limit` parameter. Together with the limit parameter, the filters make perfect sense.

lonvia:
Implemented in #742.

PR #742, opened 2023-07-25
Reverse Geocoding with osm tag

ybert:
This is an implementation of the osm tags filters for reverse geocoding.
 As proposed here #703 by @lonvia


lonvia:
Works like a charm. Merged. Thank you!

If you are in the mood to continue a bit with the code, it would be great to refactor the filter code, so that search and reverse can share the part that is now duplicated.

PR #763, opened 2023-11-19
Add status API

### LINKED ISSUE #711: Add healthcheck endpoint
It would be nice to have a `/health` or `/status` endpoint for monitoring checks.

ybert:
This is a draft about the status API feature. Resolves #711 
I'm not sure about the Nominatim part of the code for import date, so if you have any idea of improvements feel free to ask.

### DIFF
--- a/src/main/java/de/komoot/photon/App.java
+++ b/src/main/java/de/komoot/photon/App.java
@@ -207,7 +207,7 @@ public class App {
             // setup update API
             final NominatimUpdater nominatimUpdater = setupNominatimUpdater(args, server);
             get("/nominatim-update", (Request request, Response response) -> {
-                new Thread(nominatimUpdater::update).start();
+                new Thread(()-> App.startNominatimUpdate(args, server)).start();
                 return "nominatim update started (more information in console output) ...";
             });
         }
--- a/src/main/java/de/komoot/photon/App.java
+++ b/src/main/java/de/komoot/photon/App.java
@@ -12,6 +12,7 @@ import spark.Response;
 
 import java.io.FileNotFoundException;
 import java.io.IOException;
+import java.util.Date;
 
 import static spark.Spark.*;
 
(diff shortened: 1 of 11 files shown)


ybert:
@lonvia Could you review it when you have time ?

src/main/java/de/komoot/photon/App.java
```diff
@@ -162,6 +164,9 @@ private static void startApi(CommandLineArgs args, Server server) {
 
         // setup search API
         String[] langs = dbProperties.getLanguages();
+        Date importDate = dbProperties.getImportDate();
+        get("status", new StatusRequestHandler("status", importDate));
+        get("status/", new StatusRequestHandler("/status", importDate));
```
lonvia: Shouldn't the second string be also "status/" ?

src/main/java/de/komoot/photon/elasticsearch/Server.java
```diff
@@ -226,6 +229,7 @@ public void saveToDatabase(DatabaseProperties dbProperties) throws IOException
         final XContentBuilder builder = XContentFactory.jsonBuilder().startObject().startObject(BASE_FIELD)
                         .field(FIELD_VERSION, DATABASE_VERSION)
                         .field(FIELD_LANGUAGES, String.join(",", dbProperties.getLanguages()))
+                        .field(FIELD_IMPORT_DATE, dbProperties.getImportDate() instanceof Date ? String.valueOf(dbProperties.getImportDate().getTime()) : null)
```
lonvia: ES has a native Date type. I wonder if there is a point using that. At least it would spare you the string conversion here.

src/main/java/de/komoot/photon/StatusRequestHandler.java
```diff
@@ -0,0 +1,29 @@
...
+    protected StatusRequestHandler(String path, Date importDate) {
+        super(path);
+        this.importDate = importDate;
+    }
+
+    @Override
+    public String handle(Request request, Response response) {
+        final JSONObject out = new JSONObject();
+        out.put("status", "Ok");
+        out.put("import_date", this.importDate.toString());
```
lonvia: toString() seems to produce a localized string in the local timezone here. We should force UTC here and I'd be in favour of using ISO format (see https://nominatim.openstreetmap.org/status?format=json) as everywhere else in OSM.

lonvia (commented):
Looks mostly okay to me. It won't work properly with updates but they are broken anyway. So something to fix later.

lonvia:
Apologies, my PRs are now causing conflicts.

With updates working again, the status handler will need to query the database for the current import date now instead of just returning the static date read on startup. The date also needs to be updated in the database after updates are finished.

Would you be interested to adapt the code here? It would be a nice addition to the next release.

ybert:
@lonvia I made some changes to adapt the code. What do you think ?

lonvia:
Looks good to me. Just one more tiny thing: there is another route to updating at https://github.com/komoot/photon/blob/ee3ceaa55fe19ba5e8450728d62a6fc724986409/src/main/java/de/komoot/photon/App.java#L189

This also needs to use the new update function, so that the date is set.

Once you remove the 'draft' status, I'm happy to merge this.

PR #789, opened 2024-04-09
Switch build system from Maven to Gradle

lonvia:
I hope that Gradle offers a bit more flexibility in creating jars with different configurations. And even if not, it is a huge improvement in terms of readability of the build file and availability of documentation.


simonpoole:
Will give it a look ASAP.

simonpoole:
Changes look fine to me. I ran a build on my laptop, and outside of some of the tests failing (likely because of windows issues), that looked fine too.

pom.xml
```diff
@@ -1,321 +0,0 @@
...
-                                    implementation="org.apache.maven.plugins.shade.resource.ServicesResourceTransformer" />
-                                <transformer
-                                    implementation="org.apache.maven.plugins.shade.resource.ManifestResourceTransformer">
-                                    <mainClass>de.komoot.photon.App</mainClass>
-                                    <manifestEntries>
-                                        <Multi-Release>true</Multi-Release>
-                                    </manifestEntries>
-                                </transformer>
-                                <transformer
-                                    implementation="org.apache.logging.log4j.maven.plugins.shade.transformer.Log4j2PluginCacheFileTransformer" />
```
otbutz, line 198: That one is missing in the gradle version

pom.xml
```diff
@@ -1,321 +0,0 @@
...
-                        <version>0.1.0</version>
-                    </dependency>
-                </dependencies>
-                <configuration>
-                    <createDependencyReducedPom>true</createDependencyReducedPom>
-                    <filters>
-                        <filter>
-                            <artifact>*:*</artifact>
-                            <excludes>
-                                <exclude>module-info.class</exclude>
```
otbutz, line 163: Not sure if the gradle shadow plugin is doing it by default, but these excludes are otherwise missing.

otbutz (changes requested):

pom.xml
lonvia: Most of the are automatically included. The only one missing was `META-INF/versions/*/module-info.class`.

pom.xml
lonvia: The only replacement I can find is a [very old plugin](https://plugins.gradle.org/plugin/io.paulbaker.shaded-log4j2-cachefile-transformer) that hasn't been touched since 2018. Any other suggestions for replacements?

otbutz: The log4j2 support for Gradle seems to be lacking https://github.com/apache/logging-log4j-transform/issues/73

I think we can skip it as long as we don't have multiple log4j2 plugins. Otherwise we get problems with conflicting `Log4j2Plugins.dat` files.

lonvia: OK, I'll leave it then. It seems to be compiling and running as is.

Thanks for looking and thanks for the review.

lonvia: Note to self: found the proper way to solve this. Add

    transform(com.github.jengelman.gradle.plugins.shadow.transformers.Log4j2PluginsCacheFileTransformer)

to shadow step.

Goooler: `Log4j2PluginsCacheFileTransformer` may not work, I'm fixing it in https://github.com/GradleUp/shadow/pull/1175.

PR #790, opened 2024-04-16
Make maximum results configurable

karussell:
Currently the maximum results for forward and reverse search is both at 50. It might be interesting to reduce this value for forward search (to reduce maximum load) and/or increase this value for reverse search (for broader POI searches).

### DIFF
--- a/src/main/java/de/komoot/photon/query/PhotonRequestFactory.java
+++ b/src/main/java/de/komoot/photon/query/PhotonRequestFactory.java
@@ -40,8 +40,10 @@ public class PhotonRequestFactory {
 
         PhotonRequest request = new PhotonRequest(query, languageResolver.resolveRequestedLanguage(webRequest));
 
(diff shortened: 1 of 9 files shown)


src/main/java/de/komoot/photon/query/PhotonRequestFactory.java
```diff
@@ -38,7 +40,10 @@ public PhotonRequest create(Request webRequest) throws BadRequestException {
 
         PhotonRequest request = new PhotonRequest(query, languageResolver.resolveRequestedLanguage(webRequest));
 
-        request.setLimit(parseInt(webRequest, "limit"));
+        Integer limit = parseInt(webRequest, "limit");
+        if(limit != null)
+            request.setLimit(Integer.max(Integer.min(limit, maxResults), 1));
```
lonvia, line 45: Sorry, style nitpicking: please add a space after 'if' and always curly braces, even for one-liners.

PR #815, opened 2024-06-06
add support for structured queries (opensearch only)

tobiass-sdl:
this adds a new function 
```
http://localhost:2322/structured?city=berlin
```

Supported parameters are
```
"lang", "limit",  "lon", "lat", "osm_tag", "location_bias_scale", "bbox", "debug", "zoom", "layer", "countrycode", "state", "county", "city", "postcode", "district", "housenumber", "street"
```

Result format is equal to /api.

For those parameters that are also available for /api the meaning is the same. 
CountryCode has to be the ISO2 code (e.g. DE instead of "Germany"). The country code must match exactly. Fuzzy matching does not make sense for 2 letter codes. In my experience the case that address is fine except for the country code is rare compared to nonsense hits in other countries.

In order to use structured queries, the new option "-structured" has to be used when importing from Nominatim. Expect 10-20% higher file size. If you run photon on an existing index folder, photon detects automatically if the data was imported with structured query support.
For data imported without "-structured" /structured?city=berlin returns a technical error as the route is not mapped.
Structured queries are only supported for OpenSearch based photon.

### Known issues

* state information is used with low priority. This can cause issues with cities that exist in several states (e.g. "Springfield" in the US). Reason is that states are not normalized - some documents have abbreviations like "NY", other spell "New York" out. To fix this either an alias list is needed or this is normalized on import (maybe even for nominatim). 
* no fine tuning of the scores yet - would require a large test set with reliable expected results and some automated way for guessing "good" scores.
* /structured?city=some_hamlet_or_isolated_dwelling does not always work

### DIFF
--- a/app/es_embedded/src/main/java/de/komoot/photon/Server.java
+++ b/app/es_embedded/src/main/java/de/komoot/photon/Server.java
@@ -3,6 +3,7 @@ package de.komoot.photon;
 import de.komoot.photon.DatabaseProperties;
 import de.komoot.photon.Importer;
 import de.komoot.photon.Updater;
+import de.komoot.photon.searcher.StructuredSearchHandler;
 import de.komoot.photon.searcher.ReverseHandler;
 import de.komoot.photon.searcher.SearchHandler;
 import de.komoot.photon.elasticsearch.*;
@@ -168,12 +169,12 @@ public class Server {
 
     }
 
-    public DatabaseProperties recreateIndex(String[] languages, Date importDate) throws IOException {
+    public DatabaseProperties recreateIndex(String[] languages, Date importDate, boolean supportStructuredQueries) throws IOException {
         deleteIndex();
 
         loadIndexSettings().createIndex(esClient, PhotonIndex.NAME);
 
-        new IndexMapping().addLanguages(languages).putMapping(esClient, PhotonIndex.NAME, PhotonIndex.TYPE);
+        createAndPutIndexMapping(languages, supportStructuredQueries);
 
         DatabaseProperties dbProperties = new DatabaseProperties()
             .setLanguages(languages)
@@ -181,6 +182,16 @@ public class Server {
         saveToDatabase(dbProperties);
 
         return dbProperties;
+    }
+
+    private void createAndPutIndexMapping(String[] languages, boolean supportStructuredQueries)
+    {
+        if (supportStructuredQueries) {
+            throw new UnsupportedOperationException("Structured queries are not supported for elasticsearch-based Photon. Consider to use OpenSearch.");
+        }
+
+        new IndexMapping().addLanguages(languages)
+                          .putMapping(esClient, PhotonIndex.NAME, PhotonIndex.TYPE);
     }
 
     public void updateIndexSettings(String synonymFile) throws IOException {
@@ -195,9 +206,7 @@ public class Server {
         // Sanity check: legacy databases don't save the languages, so there is no way to update
         //               the mappings consistently.
         if (dbProperties.getLanguages() != null) {
-            new IndexMapping()
-                    .addLanguages(dbProperties.getLanguages())
-                    .putMapping(esClient, PhotonIndex.NAME, PhotonIndex.TYPE);
+            this.createAndPutIndexMapping(dbProperties.getLanguages(), false);
         }
     }
 
@@ -277,6 +286,10 @@ public class Server {
 
     public SearchHandler createSearchHandler(String[] languages, int queryTimeoutSec) {
         return new ElasticsearchSearchHandler(esClient, languages, queryTimeoutSec);
+    }
+
+    public StructuredSearchHandler createStructuredSearchHandler(String[] languages, int queryTimeoutSec) {
+        throw new UnsupportedOperationException("Structured queries are not supported for elasticsearch-based Photon. Consider to use OpenSearch.");
     }
 
     public ReverseHandler createReverseHandler(int queryTimeoutSec) {
--- a/app/opensearch/src/main/java/de/komoot/photon/opensearch/AddressQueryBuilder.java
+++ b/app/opensearch/src/main/java/de/komoot/photon/opensearch/AddressQueryBuilder.java
@@ -1,0 +1,275 @@
+package de.komoot.photon.opensearch;
+
+import de.komoot.photon.Constants;
+import org.apache.commons.lang3.StringUtils;
+import org.opensearch.client.opensearch._types.FieldValue;
+import org.opensearch.client.opensearch._types.query_dsl.*;
+import org.opensearch.common.unit.Fuzziness;
+
+import java.util.Objects;
+
+public class AddressQueryBuilder {
+    private static final float STATE_BOOST = 0.1f; // state is unreliable - some locations have e.g. "NY", some "New York".
+    private static final float COUNTY_BOOST = 4.0f;
+    private static final float CITY_BOOST = 3.0f;
+    private static final float POSTAL_CODE_BOOST = 7.0f;
+    private static final float DISTRICT_BOOST = 2.0f;
+    private static final float STREET_BOOST = 5.0f; // we filter streets in the wrong city / district / ... so we can use a high boost value
+    private static final float HOUSE_NUMBER_BOOST = 10.0f;
+    private static final float HOUSE_NUMBER_UNMATCHED_BOOST = 5f;
+    private static final float FACTOR_FOR_WRONG_LANGUAGE = 0.1f;
+    private final String[] languages;
+    private final String language;
+
+    private BoolQuery.Builder query = QueryBuilders.bool();
+
+    private BoolQuery.Builder cityFilter;
+
+    private boolean lenient;
+
+    public AddressQueryBuilder(boolean lenient, String language, String[] languages) {
+        this.lenient = lenient;
+        this.language = language;
+        this.languages = languages;
+    }
+
+    public Query getQuery() {
+        return query.build().toQuery();
+    }
+
+    public AddressQueryBuilder addCountryCode(String countryCode) {
+        if (countryCode == null) return this;
+
+        query.filter(QueryBuilders.term().field(Constants.COUNTRYCODE).value(FieldValue.of(countryCode.toUpperCase())).build().toQuery());
+        return this;
+    }
+
+    public AddressQueryBuilder addState(String state, boolean hasMoreDetails) {
+        if (state == null) return this;
+
+        var stateQuery = GetNameOrFieldQuery(Constants.STATE, state, STATE_BOOST, "state", hasMoreDetails);
+        query.should(stateQuery);
+        return this;
+    }
+
+    public AddressQueryBuilder addCounty(String county, boolean hasMoreDetails) {
+        if (county == null) return this;
+
+        AddNameOrFieldQuery(Constants.COUNTY, county, COUNTY_BOOST, "county", hasMoreDetails);
+        return this;
+    }
+
+    public AddressQueryBuilder addCity(String city, boolean hasDistrict, boolean hasStreet, boolean hasPostCode) {
+        if (city == null) return this;
+
+        Query combinedQuery;
+        Query nameQuery = GetFuzzyNameQueryBuilder(city, "city").boost(CITY_BOOST)
+                .build()
+                .toQuery();
+        Query fieldQuery = GetFuzzyQueryBuilder(Constants.CITY, city).boost(CITY_BOOST).build().toQuery();
+
+        if (!hasDistrict) {
+            var districtNameQuery = GetFuzzyNameQueryBuilder(city, "district").boost(0.95f * CITY_BOOST)
+                    .build()
+                    .toQuery();
+            nameQuery = QueryBuilders.bool()
+                    .should(nameQuery)
+                    .should(districtNameQuery)
+                    .minimumShouldMatch("1")
+                    .build()
+                    .toQuery();
+
+            var districtFieldQuery = GetFuzzyQueryBuilder(Constants.DISTRICT, city).boost(0.95f * CITY_BOOST).build().toQuery();
+            fieldQuery = QueryBuilders.bool()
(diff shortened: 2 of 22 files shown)


tobiass-sdl:
- [ ] adapt readme / documentation

app/opensearch/src/main/java/de/komoot/photon/opensearch/AddressQueryBuilder.java
```diff
@@ -0,0 +1,340 @@
...
+public class AddressQueryBuilder {
+    private static final float StateBoost = 0.1f; // state is unreliable - some locations have e.g. "NY", some "New York".
+    private static final float CountyBoost = 4.0f;
+    private static final float CityBoost = 3.0f;
+    private static final float PostalCodeBoost = 7.0f;
+    private static final float DistrictBoost = 2.0f;
+    private static final float StreetBoost = 5.0f; // we filter streets in the wrong city / district / ... so we can use a high boost value
+    private static final float HouseNumberBoost = 10.0f;
+    private static final float HouseNumberUnmatchedBoost = 5f;
+    private static final float FactorForWrongLanguage = 0.1f;
```
lonvia: Style nitpicking: can you make these constant upper-case.

app/opensearch/src/main/java/de/komoot/photon/opensearch/IndexMapping.java
```diff
@@ -32,12 +36,12 @@ public IndexMapping addLanguages(String[] languages) {
...
                 mappings.properties(String.format("%s.%s", field, lang),
                         b -> b.text(p -> p
-                                .index(false)
+                                .index(shouldIndexAddressField(field))
                                 .copyTo("collector.base", "collector." + lang)));
             }
 
             mappings.properties("name." + lang,
-                    b -> b.text(p -> p.index(false)
+                    b -> b.text(p -> p.index(supportStructuredQueries)
```
lonvia: Do you really need this index? Names really should be queried only through the .ngrams or .raw indexes.

src/main/java/de/komoot/photon/query/StructuredPhotonRequest.java
```diff
@@ -0,0 +1,94 @@
...
+
+    public boolean hasDistrict() { return this.district != null; }
+
+    public boolean hasCityOrPostcode() { return this.city != null && this.postCode != null; }
+
+    public boolean hasCounty() { return this.county != null; }
+
+    public boolean hasStreet() { return this.street != null || this.houseNumber != null; }
+
+    public boolean hasHouseNumber() { return this.houseNumber != null; }
```
lonvia: You  might want to extend all these to check for the empty string. (Or do the check in the set functions and set to null when the input string is empty.)

app/es_embedded/src/main/java/de/komoot/photon/Server.java
```diff
@@ -183,6 +184,16 @@ public DatabaseProperties recreateIndex(String[] languages, Date importDate) thr
         return dbProperties;
     }
 
+    private void createAndPutIndexMapping(String[] languages, boolean supportStructuredQueries)
+    {
+        if(supportStructuredQueries) {
```
lonvia: style nitpicking: please leave spaces between `if` and the bracket. (Also in other parts of the code.)

app/opensearch/src/main/java/de/komoot/photon/opensearch/AddressQueryBuilder.java
```diff
@@ -0,0 +1,340 @@
...
+                                .value(FieldValue.of(objectType))
+                                .build()
+                                .toQuery())
+                            .build()
+                            .toQuery())
+                    .boost(boost);
+        }
+    }
+
+    private void AddQuery(Query clause) {
```
lonvia: style-nitpicking: functions should start with a lower-case letter.

app/opensearch/src/main/java/de/komoot/photon/opensearch/AddressQueryBuilder.java
```diff
@@ -0,0 +1,340 @@
...
+        for(String lang : languages)
+        {
+            float boost = lang.equals(language) ? 1.0f : FactorForWrongLanguage;
+            or.should(QueryBuilders.matchPhrase()
+                    .field(name + '.' + lang)
+                    .query(value)
+                    .boost(boost)
+                    .build()
+                    .toQuery());
+        }
```
lonvia, line 262: Instead of creating a complex query over all languages here, you could also work with a collector field for each address part. Define a field 'state_collector' which is indexed but not stored. Then instead of setting the index for 'state.<lang>', you copy to 'state_collector'.

You'd loose the boost factor for wrong language. In my experiments it made little enough difference in accuracy for quite a bit of performance boost.

app/opensearch/src/main/java/de/komoot/photon/opensearch/SearchQueryBuilder.java
```diff
@@ -142,6 +145,37 @@ public SearchQueryBuilder(String query, String language, String[] languages, boo
                         .field(String.format("name.%s.raw", language))));
     }
 
+    public SearchQueryBuilder(StructuredPhotonRequest request, String language, String[] languages, boolean lenient)
+    {
+        var query4QueryBuilder = new AddressQueryBuilder(lenient, language, languages)
+                .addCountryCode(request.getCountryCode())
+                .addState(request.getState(), request.hasCounty() || request.hasCityOrPostcode())
+                .addCounty(request.getCounty(), request.hasCityOrPostcode())
```
lonvia: Those two are missing `hasStreet()` in the 'hasMoreDetails' field. It might seem that it doesn't make much sense to have 'street' and 'county' in the query but no 'city' parameter, but that has never stopped users from trying anyway.

app/opensearch/src/main/java/de/komoot/photon/opensearch/AddressQueryBuilder.java
```diff
@@ -0,0 +1,340 @@
...
+                    .should(fieldQuery.build().toQuery())
+                    .should(nameFieldQuery.filter(QueryBuilders.term()
+                                .field(Constants.OBJECT_TYPE)
+                                .value(FieldValue.of(objectType))
+                                .build()
+                                .toQuery())
+                            .build()
+                            .toQuery())
+                    .boost(boost);
+        }
```
lonvia, line 288: I'm not sure I understand the logic here. I would expect that you'd always search in the address field when hasMoreDetails is true and in the name otherwise. Right now it looks like you can also return results where all searched fields are in the address only. That seems rather confusing. For example a query with `/structured?city=vaduz` returns the city of Vaduz as expected but then also a lot of address results.

app/opensearch/src/main/java/de/komoot/photon/opensearch/AddressQueryBuilder.java
```diff
@@ -0,0 +1,340 @@
...
+                                .boost(HouseNumberUnmatchedBoost)
+                                .build()
+                                .toQuery())
+                        .should(QueryBuilders.matchPhrase()
+                                .field(Constants.HOUSENUMBER)
+                                .query(houseNumber)
+                                .boost(HouseNumberBoost)
+                                .build()
+                                .toQuery());
+            }
```
lonvia, line 181: style: bad indent

app/opensearch/src/main/java/de/komoot/photon/opensearch/IndexMapping.java
```diff
@@ -32,12 +36,12 @@ public IndexMapping addLanguages(String[] languages) {
             for (var field: ADDRESS_FIELDS) {
                 mappings.properties(String.format("%s.%s", field, lang),
                         b -> b.text(p -> p
-                                .index(false)
+                                .index(shouldIndexAddressField(field))
```
lonvia: All the new indexes simply use the default analyser. I wonder if 'analyser("index_raw")' and searchAnalyser("search") wouldn't be better here. At a minimum, it gives you German Umlaut-Folding (ö vs. oe).

You can set analysers unconditionally. They simply have no effect when the index is disabled.

lonvia (commented):
The overall structure is fine. I've added some smaller comments regarding style as well as comments regarding index structure.

My main concern is the handling of mismatches or missing address parts in the structured query. The lenient 'should' matching produces a lot of false positive results. This would be rather unexpected.

lonvia:
Sorry for creating the conflict. I found #816 while reviewing this PR. Please just rebase on master at some point. I don't mind force pushes to PR branches.

app/opensearch/src/main/java/de/komoot/photon/opensearch/AddressQueryBuilder.java
tobiass-sdl: the idea was that you still find the city if the street/housenumber is wrong (or not yet in openstreetmap) but I clearly messed something up. Returning houses or streets when no street is given is obviously a bad idea and I thought I handled that case (and failed) -> have to investigate.

app/opensearch/src/main/java/de/komoot/photon/opensearch/OpenSearchStructuredSearchHandler.java
```diff
@@ -0,0 +1,85 @@
...
+        int limit = photonRequest.getLimit();
+        int extLimit = limit > 1 ? (int) Math.round(photonRequest.getLimit() * 1.5) : 1;
+
+        var results = sendQuery(queryBuilder.buildQuery(), extLimit);
+
+        if (results.hits().total().value() == 0) {
+            results = sendQuery(buildQuery(photonRequest, true).buildQuery(), extLimit);
+
+            if (results.hits().total().value() == 0 && photonRequest.hasStreet()) {
+                var street = photonRequest.getStreet();
```
tobiass-sdl, line 47: this turned out to be easier than having an unmaintainable complex query.
Of course this could also be done on  the client side...

app/opensearch/src/main/java/de/komoot/photon/opensearch/AddressQueryBuilder.java
```diff
@@ -0,0 +1,275 @@
...
+                    .minimumShouldMatch("1")
+                    .build()
+                    .toQuery();
+        }
+
+        if (!hasStreet && !hasDistrict) {
+            // match only name
+            if (hasPostCode) {
+                // post code can implicitly specify a district that has the city in the address field (instead of the name)
+                combinedQuery = QueryBuilders.bool()
```
tobiass-sdl, line 103: e.g. /structured?state=Baden Württemberg&district=Plotzsägmühle&housenumber=2

app/opensearch/src/main/java/de/komoot/photon/opensearch/IndexMapping.java
```diff
@@ -30,10 +36,14 @@ public IndexMapping addLanguages(String[] languages) {
...
-                                .index(false)
-                                .copyTo("collector.base", "collector." + lang)));
+                var propertyName = String.format("%s.%s", field, lang);
+                var collectors = new ArrayList<>(Arrays.asList("collector.base", "collector." + lang));
+                if (shouldIndexAddressField(field)) {
+                    collectors.add(getAddressFieldCollector(field));
+                }
+
+                mappings.properties(propertyName,
+                        b -> b.text(p -> p.copyTo(collectors)));
```
lonvia: Now the 'index(false)' is gone. I'm not entirely sure what the default is, so prefer explicit statements.

app/opensearch/src/main/java/de/komoot/photon/opensearch/IndexMapping.java
```diff
@@ -91,12 +101,25 @@ private void setupBaseMappings() {
...
+                                .analyzer("index_raw"))
+                );
+
+                collectors.add(collectorName);
+            }
+
             mappings.properties(field + ".default", b -> b.text(p -> p
-                    .index(false)
-                    .copyTo("collector.default", "collector.base")));
+                    .index(shouldIndexAddressField(field))
```
lonvia: Do you still need an index here?

lonvia (commented):
The query works much better now.

You don't allow typos in the second lenient try. Is this intentional? I would have expected auto fuzziness to be enabled for it.

`/structured/?city=vaduz&street=lettstrasse` still returns me the bus stop "Lettstrasse". We might need a layer filter here.

The other thing that stands out is that the usually abbreviations don't work ('strasse' -> 'str'). The freetext search mitigates that somewhat by always doing a prefix search so that some abbreviations just accidentally work (doesn't help with 'road' -> 'rd', though). We don't have this here. I would defer this topic to later. Just wanted to mention it, in case somebody tries this and doesn't get results.

On the nitpicking side: there are still some function names that start with an upper-case letter.

tobiass-sdl:
Luckily I got rid of the bus stops without an overly complex layer filter. There was already a filter to filter out results with house numbers if the request does not contain one. Bus stops and a few other locations passed that filter because they have no house number. Therefore fixing the filter from "no house number" to "no house number and not a house" should be sufficient.
I totally agree with ignoring abbrevations for know - that's similar to the issue with state="NY" vs state="New York".
No fuzziness for lenient=false was a bug.

lonvia:
> Therefore fixing the filter from "no house number" to "no house number and not a house" should be sufficient.

You may even want to put a global filter on the query "when type is house then it must have a house number". And also add a global filter for "type != other". This should consistently filter out all object that are not "address-like".

lonvia:
Looks good as a first version. We can fine-tune in follow-up PRs, if necessary.

lonvia:
Oh dear, now I forgot about documentation. Do you mind adding a little bit of documentation for the new call in a separate PR?

I would suggest to do this in an extra file `docs/structured.md` and in the README.md just link to it (with the warning that it is optional). If the call is described directly in the main README, we will end up with repeated questions why it doesn't work on photon.komoot.io.

tobiass-sdl:
Yes, I can add some documentation. I'm off next week, not sure whether I get to it before the 20th.

Issue #259, opened 2017-10-13
Polygon not returned as Search Result


nafis042:
Whenever I search a query it always returns as point but not polygon or linestring. I can see the osm_type="R", but in geometry field always shows type="point". how can I get these polygon as geometry?

dve:
The polygon seems not to be saved in ES, according to https://github.com/komoot/photon/blob/master/src/main/java/de/komoot/photon/PhotonDoc.java.
Nominatim seems to "know" the polygon: https://github.com/openstreetmap/Nominatim/blob/master/sql/tables.sql
So it should be possible to store in ES and deliver it by the service.

Is this not done for a special reason, or did just no one needed the feature?

lonvia:
Photon only exports the centroid and the bounding box of the geometry from Nominatim. Any polygon or way geometries are lost. Adding full geometries would increase the database size quite a bit.

ducdigital:
is it possible to config Photon to load geometries in ES too?

kurko:
> Adding full geometries would increase the database size quite a bit.

Perhaps it'd make sense for an option to exist during importation so the user can decide on whether to import it or not.

GeorgeKontsevik:
I sent the geocoded coordinate to nominatim to get the polygon. This combo worked quite well for me (since the nominatim is very picky for the input address format)

Issue #357, opened 2018-09-14
Discussion for PR to improve accuracy of reverse geociding


LEDfan:
Hi

We are looking into using Photon as a (reverse) geocoding solution. The speed and simple setup is great, thanks for that :+1:!

Currently we are bumping into some inaccuracies when reverse geocoding. Especially on parts of roads where no building are nearby.

To give an example, the point lat 51.32965, lon: 4.52222, is giving weird results:
 - Nominatim retunrs:
    - Address: Essensteenweg Brasschaat Belgium
    - Matching coordinate: 51.3296452381545,4.52221194871582  
    - Link: https://nominatim.openstreetmap.org/reverse.php?format=html&lat=51.32965&lon=4.52222&zoom=16
 - Photon returns:
    - Address: Bredabaan Brasschaat Belgium
    - Matching coordinate: 51.329201, 24.5229644
    - Link: http://photon.komoot.de/reverse?lat=51.32965&lon=4.52222&distance_sort=true

The big difference between Nominatim and Photon is of course that Nominatim is using PostGIS and Photon is using ES. I did some research and I think the problem is that Nominatim is comparing the distance of an object to its full shape and Photon don't. Instead Photon only looks at some points of the object. This is not a surprise since ES doesn't support such a query (https://github.com/elastic/elasticsearch/issues/13351).

I have a possible solution that basically works as follows:
 - when indexing ES with data from Nominatim, check if the object is a highway
 - if so, divide the highway into straight parts (using the `ST_DumpPoints` PostGIS function).
 - loop over all these parts:
   - if a part is larger than 20 meters: split this part into points, where the distance between two continuously points is 20 meters

These two images explains it (the second image is indexed with 50 meters distance, but this isn't good enough).
![selection_005](https://user-images.githubusercontent.com/2996275/45557006-e9829d80-b83c-11e8-94fd-1045721505f0.png)
![selection_002](https://user-images.githubusercontent.com/2996275/45557028-f69f8c80-b83c-11e8-9d30-056c11caad84.png)

I open this issue because I want to know if you are interested in accepting a PR for this.
A POC implementation is available at https://github.com/komoot/photon/compare/master...inuits:indexed_shapes?expand=1. It needs some polishing (and maybe some tests) but it is working fine here. The same example now returns Essensteenweg.
During import from Nominatim I don't see a (significant) impact, also the ES database doesn't grow much.

Thanks in advance!

simonpoole:
[...]

PR #823, opened 2024-07-11
Add option for polygon/linestring in results

red-fenix:
At the moment, photon only returns the point of a location, and not the polygon (see https://github.com/komoot/photon/issues/259). This PR will add the option to add the polygon (i.e. geometry) to the Elasticsearch Index and a API parameter `polygon` to return said polygon. If no polygon exists, the point is returned.

WARNING: This will increase the Elasticsearch Index size! (~575GB for a Planet import).

To enable: add the command line argument `-use-geometry-column` whilst importing **and** add `&polygon=true` to the API call.


lonvia:
I haven't done a full review yet but I do have some general thoughts on the implementation:

* This really needs to be implemented for the OpenSearch version because the ES version of Photon is on its way out. Note that the OpenSearch variant does not use `mappings.json`. It defines its mapping in https://github.com/komoot/photon/blob/master/app/opensearch/src/main/java/de/komoot/photon/opensearch/IndexMapping.java.
* As long as the geometry isn't used for lookup, indexes should be disabled on the new field to save a bit of disk space.
* We already have a field `extent` which contains the bounding box. The geometry should replace this field, i.e. once full geometries are enabled, do not save extent and derive the extent from the geometry field when returning a result. We need to keep the centroid because it is not necessarily the geometric centroid of the geometry. (Note that extent was missing from the mapping specification so far. So likely it was in fact saved as a text field, which really is an oversight but can't really be fixed right now without creating an incompatible database version.)
* This would be the second optional database feature after #815. Before we become too prolific with command-line arguments, I'd lean towards adding a single parameter `-extra-db-features` which takes a list of features (right now: structured, geometries).
* I agree on the introduction of the `polygon` parameter but would prefer to disable it when the extended geometries are not available in the database. Otherwise there will be an endless stream of bug reports on photon.komoot.io, why the results return a point instead of polygon. You can save the state of the feature in the property table and load it from there on start, see #815 for an example on how to do it. This property also comes in handy during updates of the database. It would be useful to have exactly the same behaviour as on import then.
* This needs some tests for the import.

Two other considerations come in mind but they are easily deferred to follow-up PRs:

* Once we have full geometries, we'd want to use them for reverse lookup. See #357.
* It might be worth to slightly simplify the geometries before importing them, or at least make that an option. Nominatim always keeps the original OSM geometries which sometimes can have a lot more support points than necessary. Simplification might help to further reduce database size.

red-fenix:
> I haven't done a full review yet but I do have some general thoughts on the implementation:
> 
> * This really needs to be implemented for the OpenSearch version because the ES version of Photon is on its way out. Note that the OpenSearch variant does not use `mappings.json`. It defines its mapping in https://github.com/komoot/photon/blob/master/app/opensearch/src/main/java/de/komoot/photon/opensearch/IndexMapping.java.

Thanks. I will update the PR with the changes in this file soon.

One question though about the 'Elasticsearch is on it's way out': I've been planning to update the Elastic client to the [Java API](https://www.elastic.co/guide/en/elasticsearch/client/java-api-client/current/index.html) so you can use an existing Elasticsearch cluster instead of the internal one (newer versions of Elasticsearch don't support the Transport client). Is my new PR still a good idea?

> * As long as the geometry isn't used for lookup, indexes should be disabled on the new field to save a bit of disk space.

Will take this along as well.

> * This would be the second optional database feature after [add support for structured queries (opensearch only) #815](https://github.com/komoot/photon/pull/815). Before we become too prolific with command-line arguments, I'd lean towards adding a single parameter `-extra-db-features` which takes a list of features (right now: structured, geometries).

Agreed

> * I agree on the introduction of the `polygon` parameter but would prefer to disable it when the extended geometries are not available in the database. Otherwise there will be an endless stream of bug reports on photon.komoot.io, why the results return a point instead of polygon. You can save the state of the feature in the property table and load it from there on start, see [add support for structured queries (opensearch only) #815](https://github.com/komoot/photon/pull/815) for an example on how to do it. This property also comes in handy during updates of the database. It would be useful to have exactly the same behavior as on import then.

OK. I will make the default to return the polygon when it's available in the index. The option 'polygon=false' will return the centroid instead.

> * This needs some tests for the import.

Will do.

> 
> Two other considerations come in mind but they are easily deferred to follow-up PRs:
> 
> * Once we have full geometries, we'd want to use them for reverse lookup. See [Discussion for PR to improve accuracy of reverse geocoding #357](https://github.com/komoot/photon/issues/357).

I have to look into this issue.

> * It might be worth to slightly simplify the geometries before importing them, or at least make that an option. Nominatim always keeps the original OSM geometries which sometimes can have a lot more support points than necessary. Simplification might help to further reduce database size.

I will look into this. 
Another thing related to this: if you're searching for a street Nominatim (and thus Photon) returns a street in parts because they are separate OSM id's.  I'm still looking for a method to merge multiple ways (i.e. linestrings) into 1 linestring to make sure the whole street is displayed instead of a part [(example)](https://nominatim.openstreetmap.org/ui/search.html?q=Leidseplein+amsterdam)

lonvia:
> One question though about the 'Elasticsearch is on it's way out': I've been planning to update the Elastic client to the [Java API](https://www.elastic.co/guide/en/elasticsearch/client/java-api-client/current/index.html) so you can use an existing Elasticsearch cluster instead of the internal one (newer versions of Elasticsearch don't support the Transport client). Is my new PR still a good idea?

We'll drop ES support completely and go with OpenSearch. Note that the OS version already supports an external OpenSearch cluster. The support is just somewhat rudimentary and HTTP-only.

red-fenix:
> Would this PR return a Multipolygon in case for queries like "hamburg"? See [the nominatim response](https://nominatim.openstreetmap.org/ui/search.html?q=hamburg).
> 
> Currently photon returns a slightly confusing, but correct extent as only a single extent is allowed https://photon.komoot.io/api/?q=hamburg ... or maybe we add a new field `extents` in the response?

Yes, it would return a Polygon like Nominatim does.

lonvia:
Sorry for the long delay in responding.

I've decided to get a release out first before looking into this again. Because of this there is a minor merge conflict now regarding the dependencies. Also, the opensearch variant doesn't compile because `setUpESWithPolygons()` is only defined for the elasticsearch variant. Could you quickly fix both issues?

src/main/java/de/komoot/photon/SearchRequestHandler.java
```diff
@@ -56,6 +58,15 @@ public String handle(Request request, Response response) {
...
         }
 
-        return new GeocodeJsonFormatter(photonRequest.getDebug(), photonRequest.getLanguage()).convert(results, debugInfo);
+        boolean returnPolygon = supportPolygons;
+        if (photonRequest.isPolygonRequest()) {
+            returnPolygon = photonRequest.getReturnPolygon();
+        }
+
+        if (!supportPolygons && (photonRequest.isPolygonRequest() && photonRequest.getReturnPolygon())) {
+            throw new BadRequestException(400, "You're requesting a polygon, but polygons are not imported!");
```
lonvia: You need to throw a `halt` at this point and make sure the content is json formatted, see line 43 above.

src/main/java/de/komoot/photon/SearchRequestHandler.java
```diff
@@ -56,6 +58,15 @@ public String handle(Request request, Response response) {
             debugInfo = requestHandler.dumpQuery(photonRequest);
         }
 
-        return new GeocodeJsonFormatter(photonRequest.getDebug(), photonRequest.getLanguage()).convert(results, debugInfo);
+        boolean returnPolygon = supportPolygons;
+        if (photonRequest.isPolygonRequest()) {
+            returnPolygon = photonRequest.getReturnPolygon();
+        }
+
+        if (!supportPolygons && (photonRequest.isPolygonRequest() && photonRequest.getReturnPolygon())) {
```
lonvia: This check comes to late. Parameter need to be tested and rejected before the actual search is done (around line 45).

src/main/java/de/komoot/photon/StructuredSearchRequestHandler.java
```diff
@@ -51,6 +51,6 @@ public String handle(Request request, Response response) {
             debugInfo = requestHandler.dumpQuery(photonRequest);
         }
  */
-        return new GeocodeJsonFormatter(photonRequest.getDebug(), photonRequest.getLanguage()).convert(results, debugInfo);
```
lonvia, line 50: I'd expect the same code here as in the SearchRequestHandler (check if available etc.)

src/main/java/de/komoot/photon/ReverseSearchRequestHandler.java
```diff
@@ -53,6 +53,6 @@ public String handle(Request request, Response response) {
             debugInfo = requestHandler.dumpQuery(photonRequest);
         }
 
-        return new GeocodeJsonFormatter(false, photonRequest.getLanguage()).convert(results, debugInfo);
+        return new GeocodeJsonFormatter(false, photonRequest.getLanguage(), photonRequest.getPolygon()).convert(results, debugInfo);
```
lonvia: I'd expect the same code here as in the SearchRequestHandler (check if available etc.)

src/main/java/de/komoot/photon/nominatim/NominatimConnector.java
```diff
@@ -24,7 +24,7 @@
 public class NominatimConnector {
     private static final Logger LOGGER = org.slf4j.LoggerFactory.getLogger(NominatimConnector.class);
 
-    private static final String SELECT_COLS_PLACEX = "SELECT place_id, osm_type, osm_id, class, type, name, postcode, address, extratags, ST_Envelope(geometry) AS bbox, parent_place_id, linked_place_id, rank_address, rank_search, importance, country_code, centroid";
+    private static final String SELECT_COLS_PLACEX = "SELECT place_id, osm_type, osm_id, class, type, name, postcode, address, extratags, ST_Envelope(geometry) AS bbox, parent_place_id, linked_place_id, rank_address, rank_search, importance, country_code, centroid, geometry";
```
lonvia: Geometries can become very large, so should only be added to the returned columns when really needed.

src/main/java/de/komoot/photon/query/PhotonRequestFactory.java
```diff
@@ -110,6 +111,11 @@ private void addCommonParameters(Request webRequest, PhotonRequestBase request)
         if (layerFiltersQueryMap.hasValue()) {
             request.setLayerFilter(layerParamValidator.validate(layerFiltersQueryMap.values()));
         }
+
+        if (webRequest.queryParams("polygon") != null) {
+            request.setPolygonRequest(true);
+            request.setReturnPolygon(Boolean.parseBoolean(webRequest.queryParams("polygon")));
```
lonvia: We'd need a full parsing function here that throws on unknown values. Question is if we should allow 0/1, yes/no besides true/false.

lonvia (commented):
Couple of more comments after trying this out. The OpenSearch version currently doesn't work at all. It already fails to import.

The output for the ElasticSearch version displays a wrong projection:

```
...
"geometry": {
  "coordinates":[9.5227103,47.1395576],
  "type":"Point",
  "crs":{"type":"name","properties":{"name":"EPSG:0"}}
}
...
```

Photon doesn't supply a CRS at all with the geometry. I'd leave it at that.

red-fenix:
> Sorry for the long delay in responding.
> 
> I've decided to get a release out first before looking into this again. Because of this there is a minor merge conflict now regarding the dependencies. Also, the opensearch variant doesn't compile because `setUpESWithPolygons()` is only defined for the elasticsearch variant. Could you quickly fix both issues?

Will do this weekend.

red-fenix:
> Couple of more comments after trying this out. The OpenSearch version currently doesn't work at all. It already fails to import.

I've made some adjustments and have been trying to get the OpenSearch version running however, it fails with an error:

`Request failed: [mapper_parsing_exception] No handler for type [geo_shape] declared on field [geometry]`

I've searching online, but can't seem to find the reason for [OpenSearch clearly has geo_shape support](https://opensearch.org/docs/latest/field-types/supported-field-types/geo-shape/).

In other news: I fixed the PR with your suggestions, but will perform some additional tests this weekend.

lonvia:
> Request failed: [mapper_parsing_exception] No handler for type [geo_shape] declared on field [geometry]

That's where I gave up my experiments with this branch as well. It's possible that the [embedded version of OpenSearch](https://github.com/codelibs/opensearch-runner/) that we are using is missing a dependency needed for geo_shape. Maybe try running against a stand-alone OpenSearch and see if it works then?

otbutz:
https://github.com/opensearch-project/geospatial ? Maybe it's enough to add the OpenSearch plugin as a dependency.

red-fenix:
So, I've found out that there were multiple issues with my build:
1. The Opensearch implementation is very different from the Elasticsearch implementation (ser/des vs parsing a Searchit), but I've made it work.

2. The main problem is the `opensearch-runner`. It doesn't have all Opensearch plugins available and the geo plugin is [disabled](https://github.com/codelibs/opensearch-runner/blob/5e236d713040a5383473c80bc7c75beb811ea6ee/src/main/java/org/codelibs/opensearch/runner/OpenSearchRunner.java#L124). I've been trying to add the Opensearch geospatial plugin as either a module or a plugin, but was unable to do so.

Therefore I switched to [testcontainers](https://github.com/opensearch-project/opensearch-testcontainers) and everything is working-ish. The unit tests in the IDE all work, but on compiling some unit tests fail.

Is there anyone who can help me add the Geo plugin to the opensearch-runner of would you be willing to switch to testcontainers and docker?

otbutz:
You could open an issue at https://github.com/codelibs/opensearch-runner. Either it's already working or it's a valid feature request.

lonvia:
It needs to work with opensearch-runner because that is also what Photon uses in embedded mode.

From the code you cite it looks like you can supply your own module list by adding `.moduleTypes(<comma-separated list of modules>)` at https://github.com/komoot/photon/blob/433839d055cb41d91566d153f8d79ff06984476f/app/opensearch/src/main/java/de/komoot/photon/Server.java#L81

lonvia:
With #855 merged you can now simply add the plugin to `build.gradle` and enable it by adding it to the list in https://github.com/komoot/photon/blob/919784fe80680631a56471b14e4648683ed568fe/app/opensearch/src/main/java/de/komoot/photon/Server.java#L36

red-fenix:
> With #855 merged you can now simply add the plugin to `build.gradle` and enable it by adding it to the list in
> 
> https://github.com/komoot/photon/blob/919784fe80680631a56471b14e4648683ed568fe/app/opensearch/src/main/java/de/komoot/photon/Server.java#L36

Thank you so far for your help. I've been trying to get it working, but failed so far. I've opened a [issue](https://github.com/codelibs/opensearch-runner/issues/3) with the Codelab Opensearch Runner.

red-fenix:
Opensearch [has fixed it](https://github.com/codelibs/opensearch-runner/issues/3) and I've updated the PR. Both Opensearch tests and Es Embedded are 100% working.

app/es_embedded/src/main/java/de/komoot/photon/Server.java
```diff
@@ -276,11 +282,15 @@ public DatabaseProperties loadFromDatabase() {
...
         String importDateString = properties.get(FIELD_IMPORT_DATE);
 
+        String supportPolygons = properties.get(FIELD_SUPPORT_POLYGONS);
+
         return new DatabaseProperties(langString == null ? null : langString.split(","),
-                                      importDateString == null ? null : Date.from(Instant.parse(importDateString)),
-                                      false);
+                importDateString == null ? null : Date.from(Instant.parse(importDateString)),
+                false,
+                supportPolygons == null ? null : Boolean.parseBoolean(supportPolygons));
```
lonvia: Just call `parseBoolean()`. It will do the right thing when `supportPolygons == null`.

app/es_embedded/src/test/java/de/komoot/photon/ESBaseTester.java
```diff
@@ -45,17 +47,21 @@ public void tearDown() throws IOException {
     }
 
     public void setUpES() throws IOException {
-        setUpES(dataDirectory, "en");
+        setUpES(dataDirectory, false,"en");
```
lonvia, line 50: Style: add space after commas (also below)

app/opensearch/src/main/java/de/komoot/photon/opensearch/PhotonDocSerializer.java
```diff
@@ -48,6 +51,40 @@ public void serialize(PhotonDoc value, JsonGenerator gen, SerializerProvider pro
...
+
+            gen.writeEndArray();
+            gen.writeEndObject();
+
+//                gen.writeObjectFieldStart("crs");
+//                gen.writeStringField("type", "name");
+//                gen.writeObjectFieldStart("properties");
+//                gen.writeStringField("name", "EPSG:0");
+//                gen.writeEndObject();
+//            gen.writeEndObject();
```
lonvia: You can't just assume that the geometry is of type polygon. There may be pretty much anything in there: point, linestring, multilinestring, polygon, multipolygon.

Looks like we need tests for all the different geometry types to make sure this works.

app/opensearch/src/main/java/de/komoot/photon/Server.java
```diff
@@ -87,6 +89,7 @@ private HttpHost[] startInternal(String clusterName) {
                 .clusterName(clusterName)
                 .numOfNode(1)
                 .moduleTypes(OPENSEARCH_MODULES)
+                .pluginTypes(OPENSEARCH_MODULES)
```
lonvia: What does this line do? I had to remove it to make the server work as it creates an exception `OpenSearchRunnerException: org.opensearch.geospatial.plugin.GeospatialPlugin is not found`

app/opensearch/src/test/java/de/komoot/photon/ESBaseTester.java
```diff
@@ -1,18 +1,23 @@
 package de.komoot.photon;
 
+import de.komoot.photon.opensearch.Importer;
 import de.komoot.photon.opensearch.OpenSearchTestServer;
+import de.komoot.photon.opensearch.Updater;
```
lonvia: Remove these two imports again, they confuse the `make...()` functions below. Once removed, the explicit casts can go away again.

src/main/java/de/komoot/photon/SearchRequestHandler.java
```diff
@@ -20,16 +20,20 @@
 public class SearchRequestHandler extends RouteImpl {
     private final PhotonRequestFactory photonRequestFactory;
     private final SearchHandler requestHandler;
+    private final int maxResults;
```
lonvia: This looks to be a merging artifact.

src/main/java/de/komoot/photon/nominatim/NominatimUpdater.java
```diff
@@ -75,16 +75,18 @@ public class NominatimUpdater extends NominatimConnector {
     // One-item cache for address terms. Speeds up processing of rank 30 objects.
     private long parentPlaceId = -1;
     private List<AddressRow> parentTerms = null;
+    private boolean useGeometryColumn;
 
 
-    public NominatimUpdater(String host, int port, String database, String username, String password) {
-        this(host, port, database, username, password, new PostgisDataAdapter());
+    public NominatimUpdater(String host, int port, String database, String username, String passwordboolean, boolean useGeometryColumn) {
+        this(host, port, database, username, passwordboolean, new PostgisDataAdapter(), useGeometryColumn);
```
lonvia: Change 'passwordboolean' seems unrelated.

src/main/java/de/komoot/photon/query/PhotonRequestBase.java
```diff
@@ -18,6 +18,8 @@ public class PhotonRequestBase
     private int zoom = 14;
     private Envelope bbox = null;
     private boolean debug = false;
+    private boolean returnPolygon = true;
+    private boolean polygonRequest = false;
```
lonvia: Why do you need two parameters here?

src/main/java/de/komoot/photon/query/PhotonRequestFactory.java
```diff
@@ -145,4 +151,15 @@ private Double parseDouble(Request webRequest, String param) throws BadRequestEx
 
         return outVal;
     }
+
+    private Boolean parseBoolean(Request webRequest, String param) throws BadRequestException {
```
lonvia: remove `throws`

lonvia (commented):
Still not completely finished with a detailed review but I added some comments.

I currently see two major issues that still need to be solved:
* the OS variant is currently hard-coded towards polygons with the result that streets cannot be found anymore. The code should be able to deal with any geometries including invalid ones.
* import in the ES variant with geometry enabled is very slow for me. Slower by a factor of 5. Any idea what is causing this?

The first issue does make me wonder if we shouldn't rather call the API parameter `geometry` and consistently go with that in the code. "polygon" is quite confusing for something that might also cause linestrings to appear. We could then have `geometry=centroid/bbox/full`.

app/es_embedded/src/main/java/de/komoot/photon/Server.java
red-fenix: Fixed

app/es_embedded/src/test/java/de/komoot/photon/ESBaseTester.java
red-fenix: Fixed

app/opensearch/src/main/java/de/komoot/photon/opensearch/PhotonDocSerializer.java
red-fenix: I've added support for LineString and Polygons for now. I have to examine the Nominatim database if they use other formats.

app/opensearch/src/main/java/de/komoot/photon/Server.java
red-fenix: Removed legacy code.

src/main/java/de/komoot/photon/SearchRequestHandler.java
red-fenix: Fixed

src/main/java/de/komoot/photon/nominatim/NominatimUpdater.java
red-fenix: Fixed

src/main/java/de/komoot/photon/query/PhotonRequestBase.java
red-fenix: Once upon a time, I've had my reasons. Can't remember what they were, so I removed it. If 'geometry' is supported by the database, the default is that they will be returned.

src/main/java/de/komoot/photon/query/PhotonRequestFactory.java
red-fenix: Done

app/opensearch/src/test/java/de/komoot/photon/ESBaseTester.java
red-fenix: Done

red-fenix:
> Still not completely finished with a detailed review but I added some comments.
> 
> I currently see two major issues that still need to be solved:
> 
> * the OS variant is currently hard-coded towards polygons with the result that streets cannot be found anymore. The code should be able to deal with any geometries including invalid ones.

I have build in support for LineString and Polygons and have to analyze the Nominatim database for other Geometries.

> * import in the ES variant with geometry enabled is very slow for me. Slower by a factor of 5. Any idea what is causing this?
I will look into this, but since Elasticsearch support will be deprecated, how much time do I need to spend on this?

> 
> The first issue does make me wonder if we shouldn't rather call the API parameter `geometry` and consistently go with that in the code. "polygon" is quite confusing for something that might also cause linestrings to appear. We could then have `geometry=centroid/bbox/full`.

I will rename it to 'geometry' support in my next commit.

red-fenix:
Another thing: the Tests are working for Elastisearch as well as Opensearch. Though, whilst importing data, I get an 'Error during bulk import.'

Debugging this from the commandline is very hard, because I can't see the actual data that is being imported. I've been unable to run the import process in debugging from IntelliJ (I keep getting the error 'Jar hell'). 

Is there any way I can run the import process from Intellij so that I can better debug the error?

lonvia:
> I have build in support for LineString and Polygons and have to analyze the Nominatim database for other Geometries.

Nominatim has everything: Point, LineString, MultiLineString, Polygon, MultiPolygon. It definitely would be better to use a geojson builder instead of hand-coding the conversion. I see that you did that for ES. I do wonder though if that is the part that is making ES slow. That is why I was asking about the performance.

> Is there any way I can run the import process from Intellij so that I can better debug the error?

Not sure where the jar hell comes from. I've only seen this in the ES variant. Maybe you could look into improving error reporting here? [This code](https://github.com/komoot/photon/blob/919784fe80680631a56471b14e4648683ed568fe/app/opensearch/src/main/java/de/komoot/photon/opensearch/Importer.java#L57) really could have a more helpful message and print the errors that the bulk import returned.

red-fenix:
OK. I've finally been able to start fixing the mentioned issues. For anyone wanting to run Photon in Intellij, create a new _run configuration_ based on the `photon [run]` (app:opensearch:run). 

In the `Run` field you can add the photon arguments behind `app:opensearch:run` i.e.: `--args="-nominatim-import -import-geometry-column -password very_secure_password"`

app/opensearch/src/main/java/de/komoot/photon/opensearch/IndexMapping.java
```diff
@@ -78,6 +78,7 @@ private void setupBaseMappings() {
         }
 
         mappings.properties("coordinate", b -> b.geoPoint(p -> p));
+        mappings.properties("geometry", b -> b.geoShape(p -> p));
```
lonvia, line 81: I think this will create an index on "geometry" by default. We don't need that.

src/main/java/de/komoot/photon/PhotonDoc.java
```diff
@@ -44,6 +45,15 @@ public PhotonDoc(long placeId, String osmType, long osmId, String tagKey, String
         this.tagValue = tagValue;
     }
 
+    public PhotonDoc(long placeId, String osmType, long osmId, String tagKey, String tagValue, Geometry geometry) {
```
lonvia: No need, to have an extra variant of a constructor here. `PhotonDoc(...).geometry(whatever)` will have the same effect.

lonvia (commented):
What is the status here? It looks to be shaping up quite nicely now. Did a quick import test and the performance for OS is okay. The main major change would be a general shift from "polygon" to "geometry" in naming. I really think that will make life easier for future maintainers. 

A smaller thing: the output still has `"crs":{"type":"name","properties":{"name":"EPSG:0"}}` in there. Either get rid of the crs completely or see if we can force the correct EPSG 4326, preferably, on import already.

I'm also still not entirely happy with the proliferation of command line parameters but I think it is better to improve that in a separate refactoring PR after this one.

There are a number of nitpicking comments (a couple of formatting issues etc.) but it is usually better to do those only at the end.

app/opensearch/src/main/java/de/komoot/photon/opensearch/IndexMapping.java
red-fenix: Unfortunately there is no option to apply ```index(false)``` in the geo_shape field.

red-fenix:
I've processed your latest remarks. Let me know what you think!

lonvia:
I'm inclined to merge this now as is. There are a few places where there is room for improvement (for example, there could be more tests) but given the size of the PR it will be easier to do those in separate PRs.

Could you rebase your branch on current master and squash the commits together in a handful of meaningful commits? Ideally unrelated changes like the improved [error output here](https://github.com/komoot/photon/pull/823/files#diff-770a4dc2b5bc41538fb248f1b12e2e61a13caae522744d5dc178f918d9b490baR58) will go into their own commits. But if it is too much to sort out in hindsight, one single huge commit will do as well.

lonvia:
I've squash-merged this now as it is blocking a few other changes and I really don't want to make you go through rebasing again. There probably will be a few minor changes on top.

Thanks for all your work on this and for patiently going through all the remarks.

Issue #213, opened 2016-02-03
Reverse geocoding radius


vera20:
How I can increase the search radius or limit result?

Svantulden:
You could change the distance on this line: https://github.com/komoot/photon/blob/master/src/main/java/de/komoot/photon/query/ReverseQueryBuilder.java#L112

vera20:
How I can get more results in reverse geocoding? I change the distance radius but I still get one result

Svantulden:
I think you can change this line (https://github.com/komoot/photon/blob/master/src/main/java/de/komoot/photon/ReverseSearchRequestHandler.java#L50) to get more results.

masda70:
This pull should take care of this issue.

https://github.com/komoot/photon/pull/219

karussell:
Fixed via #254

PR #825, opened 2024-07-31
Add note to readme about radius parameter for reverse geocoding

weskamm:
Found this parameter by searching through issues (https://github.com/komoot/photon/issues/213) , and added documentation for it


lonvia:
Thank you.

PR #827, opened 2024-08-15
Allow multiple CORS origins to be specified

burleight:
We'd find it useful if we could specify multiple, comma separated origins when using the `--cors-origin` flag

### DIFF
--- a/src/test/java/de/komoot/photon/ApiIntegrationTest.java
+++ b/src/test/java/de/komoot/photon/ApiIntegrationTest.java
@@ -10,6 +10,7 @@ import java.io.BufferedReader;
 import java.io.InputStreamReader;
 import java.net.HttpURLConnection;
 import java.net.URL;
+import java.net.URLConnection;
 import java.util.stream.Collectors;
 
 import static org.junit.jupiter.api.Assertions.*;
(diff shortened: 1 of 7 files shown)


src/test/java/de/komoot/photon/ApiIntegrationTest.java
```diff
@@ -72,6 +73,57 @@ void testCorsOriginIsSetToSpecificDomain() throws Exception {
...
+        try {
+            // allows us to set the Origin header
+            System.setProperty("sun.net.http.allowRestrictedHeaders", "true");
+
+            String[] origins = {"www.randomsite.com", "www.arbitrary.com"};
+            for (String origin: origins) {
+                URLConnection urlConnection = new URL("http://127.0.0.1:" + port() + "/api?q=berlin").openConnection();
+                urlConnection.setRequestProperty("Origin", origin);
+                HttpURLConnection connection = (HttpURLConnection) urlConnection;
+                assertNotEquals(origin, connection.getHeaderField("Access-Control-Allow-Origin"));
```
lonvia: If I read the code right, this should return the first configured origin. So better test for `assertEquals("www.poole.ch", ...`.

lonvia (commented):
Looks good to me in principle. Clean code and the tests are all there.

Not sure why the tests don't pass.

src/test/java/de/komoot/photon/ApiIntegrationTest.java
burleight: Thanks for the feedback.

I've made this change, and fixed the tests with

```
test {
    systemProperty "sun.net.http.allowRestrictedHeaders", "true"
}
```

burleight:
@lonvia  - hopefully this third try does it. I've moved the test configuration into `shared.gradle`

lonvia:
It did indeed. Thank you!

PR #874, opened 2025-03-14
README: point people to forum and discussions instead of mailing list

lonvia:
Point people to Github's discussion feature and the OpenStreetMap community forum for discussion.

The Photon mailinglist has seen very little activity in the last years, just like most other mailing lists. So the OSM administrators are considering to shut down the mailing list service. Lets help them and move on.


otbutz:
Could you also add an issue template like the one in the Nominatim repo?

grafik

Issue #900, opened 2025-06-01
Photon request is incomplete


hircumg:
Hello,
I'm trying to use photon instead of Nominatim for better features search but I've got sucked on the following issue:

- When I was tried to search a street I've got only part of in, how to fix this?

Example:
- Zagreb, Ulica Dore Pfanove

In OSM there is two parts of it:

![Image](https://github.com/user-attachments/assets/9fc0e88a-ee50-4659-944b-44d883c4e232)

The photon returns only one part of it (osm id 364140392):
- Request https://photon.komoot.io/api?q=%22%20Zagreb,%20Ulica%20Dore%20Pfanove%22

From the other side, reverse request to photon could extract second part of the street (osm id 78500211):
- Request https://photon.komoot.io/reverse?lat=45.803396&lon=15.941833

I've also tried to add limit, layer=street or update the request but nothing helped.

PR #903, opened 2025-06-07
Added dedupe arg for requests

hircumg:
After adding geometries to response dedupling might filter our useful data, e.g. #900.

The idea is to add `dedupe` argument for requests which will disable StreetDupesRemover and response all results.

### DIFF
--- a/src/main/java/de/komoot/photon/query/RequestBase.java
+++ b/src/main/java/de/komoot/photon/query/RequestBase.java
@@ -8,6 +8,7 @@ public class RequestBase {
     private String language = "default";
     private int limit = 15;
     private boolean debug = false;
+    private boolean dedupe = true;
     private boolean returnGeometry = false;
 
     private final List<TagFilter> osmTagFilters = new ArrayList<>(1);
@@ -23,6 +24,10 @@ public class RequestBase {
 
     public boolean getDebug() {
         return debug;
+    }
+
+    public boolean getDedupe() {
+        return dedupe;
     }
 
     public boolean getReturnGeometry() {
@@ -52,6 +57,12 @@ public class RequestBase {
     public void setDebug(Boolean debug) {
         if (debug != null) {
             this.debug = debug;
+        }
+    }
+
+    public void setDedupe(Boolean dedupe) {
+        if (dedupe != null) {
+            this.dedupe = dedupe;
         }
     }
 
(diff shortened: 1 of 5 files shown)


src/main/java/de/komoot/photon/query/RequestBase.java
```diff
@@ -8,6 +8,7 @@ public class RequestBase {
     private String language = "default";
     private int limit = 15;
     private boolean debug = false;
+    private boolean dedupe = true;
```
lonvia, line 11: The parameter makes only sense for forward search, so the code from this file should go into [SearchRequestBase.java](https://github.com/komoot/photon/blob/master/src/main/java/de/komoot/photon/query/SearchRequestBase.java)

src/main/java/de/komoot/photon/query/RequestFactoryBase.java
```diff
@@ -12,7 +12,7 @@
 
 public class RequestFactoryBase {
     protected static final Set<String> BASE_PARAMETERS = Set.of(
-            "lang", "limit", "debug", "geometry", "osm_tag", "layer");
+        "lang", "limit", "debug", "dedupe", "geometry", "osm_tag", "layer");
```
lonvia, line 15: dto here, the right place would be `SearchRequestFactoryBase.java`

lonvia (commented):
I'm in general happy to merge this. Apart from moving the code, would you mind adding a small test for this. None of the existing tests has the appropriate test data, I'm afraid, so you need to create your own. Modelling after https://github.com/komoot/photon/blob/master/src/test/java/de/komoot/photon/query/QueryBasicSearchTest.java or https://github.com/komoot/photon/blob/master/src/test/java/de/komoot/photon/api/ApiIntegrationTest.java will both work.

src/main/java/de/komoot/photon/query/RequestBase.java
hircumg: @lonvia, ok, I'll move it. The reason why I added this parameter globally because StreetDupesRemover used along all requests

lonvia: Oooh, I changed that recently when deduplicating the code. You are right. Ignore me then and leave the code as is. Sorry about the noise.

hircumg: sure, in this case I will only add some tests

hircumg:
@lonvia I've added a small test as you proposed. Please, let me know if you would like to add more.

lonvia:
Looks good to me this way. Thanks.

PR #909, opened 2025-06-11
Instructions about how to update the database from graphopper

laem:


lonvia:
Sorry, took me a while to decide if the README is really the right place for that. Given that photon does target people with little admin experience, it is probably a hint worthwhile having.

Thanks.

PR #916, opened 2025-06-27
Switch to higher compression for OpenSearch

lonvia:
Switch to [best_compression](https://docs.opensearch.org/docs/2.19/im-plugin/index-codecs/) for index compression. This saves around 10% (or 20GB for the planet) on disk space. OpenSearch is supposedly able to work with zstd as well but that does not seem to be available with opensearch-runner.


otbutz:
Any exceptions with `zstd` ? It might be worth to raise that as an issue.

lonvia:
'zstd' is simply not in the list of available codecs. The additional gain with zstd is not sufficient to justify additional time on my side for investigating the issue. I do welcome help on that, though.

PR #917, opened 2025-06-27
Correctly handle addr:street and addr:place appearing together

### LINKED ISSUE #914: Addresses with addr:street and addr:place not handled correctly
When addr:street and addr:place appear in the same object, then addr:street should take precedence. Currently addr:place will overwrite the street name.

lonvia:
Give preference to `addr:street` when it appear together with `addr:place`. When they do appear together then `addr:place` is usually a synonym for `addr:neighbourhood`. The only exception is when there is a street or conscription number but that case is already covered.

`addr:block_number` keeps its priority over `addr:street`. There are very few cases, where they appear together and these look like mis-taggings on the `addr:street` side. Might be worth investigating, if we ignore block numbers outside Japan. I'm not sure there is any other country where the use is actual valid.

Fixes #914.


lonvia:
AI hallucinations. There is no such thing as a `addConscriptionAddress` as far as I can see.

PR #926, opened 2025-07-18
Improvments for startup and shutdown

lonvia:
On startup, Photon will now only create the database directory when in import mode. In all other mode if expects the directory to be existing and will bail out if it cannot be found. This should make it easier for users to understand the error when they have downloaded the wrong index dump (ElasticSearch vs. OpenSearch). OpenSearch makes an exception when using an external OS instance. The data directory will be ignored in this case. The situation is slightly different for ElasticSearch because it still needs to copy in the painless plugin.

On shutdown, Photon now catches when the API is stopped via Ctrl-C or similar and properly shuts down the node first when running with an internally started database. I've seen very rare cases of database corruption in the past. Proper shutdown hopefully solves those.

### DIFF
--- a/app/es_embedded/src/main/java/de/komoot/photon/Server.java
+++ b/app/es_embedded/src/main/java/de/komoot/photon/Server.java
@@ -80,7 +80,7 @@ public class Server {
         }
     }
 
-    public Server start(String clusterName, String[] transportAddresses) {
+    public void start(String clusterName, String[] transportAddresses) {
         Settings.Builder sBuilder = Settings.builder();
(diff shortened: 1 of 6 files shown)


lonvia:
Sure. And when we don't do this else-if, we have potentially esServer.close() being called concurrently from two different threads. Fun, isn't it?

lonvia:
Alright, version with proper thread-safe shutdown of `esServer`. There might still be some odd errors in the log messages if Javalin is trying to server last requests while esServer is already in shutdown mode but that should be recoverable.

Issue #523, opened 2020-12-30
Requirements are missing in readme


vguleria1108:
I am trying to deploy photon on elasticbean but my instance keeps stopping, so if you can add minimum requirements required than it will be helpful.

kenseii:
Hi @vguleria1108 did you find the answer for this? i am trying to deploy photon as well

lonvia:
The instance at photon.komoot.io currently runs at a machine with 64GB RAM and 6 CPUs. But that's on the upper end of what you need. 32GB RAM will do for less busy installations, maybe even 16GB are enough. I have no reliable numbers for minimum requirements which is why the ticket is still open.

This is all for running a planet-wide photon from the pre-computed data dumps we provide. If you want to create the database yourself, the [hardware requirements for Nominatim](https://nominatim.org/release-docs/latest/admin/Installation/#hardware) apply.

akatie:
The deployment was successful.

lonvia:
On what kind of hardware?

otbutz:
@lonvia we're running Photon successfully with a Europe dataset using `-Xms2G -Xmx2G` on VMs with at least 4 cores and 12GB RAM. For the duration of the import we're using `-Xms5G -Xmx5G`.

henrik242:
Are there any plans on raising the JDK requirement to the latest LTS (21)?

lonvia:
No. Policy is to support the three latest LTS unless there is a requirement from the dependencies that cannot be worked around. So, we probably raise to 17 by end of year but certainly not to 21.

henrik242:
I have usually been able to get newer jdks on older distros without issues, though.

lonvia:
Overhauled the README. The recommendations remain vague but that's as godd as it gets because requirements depend a lot on the kind of load the service gets.

PR #928, opened 2025-07-31
Remove ElasticSearch version

lonvia:
This gets rid of the ElasticSearch backend, making OpenSearch the only available storage option. Simplifies the gradle build files accordingly.

Also updates dependencies to their latest version.


otbutz:
I don't know if ES was a blocker, but since we're already sunsetting things, could we bump the Java version?

lonvia:
See https://github.com/komoot/photon/issues/523#issuecomment-2866700184 ff

otbutz:
https://github.com/search?q=repo%3Akomoot%2Fphoton%20elastic&type=code

Looks like there are still a few remnants of the old backend.

PR #908, opened 2025-06-11
Add filtering for osm tags ("extra" properties)

henrik242:
Example query: `/api?q=Oslo&osm_tag=extra.transport_modes:onstreetTram`

The extra tag(s) must have been imported, e.g.:

    $ java -jar target/photon-opensearch-0.7.0.jar -nominatim-import \
           -import-file oslo.ndjson -languages no,en -extra-tags transport_modes

PR #948, opened 2025-10-02
[WIP] Add customizable filterable categories

lonvia:
This PR officially introduces _categories_ to Photon. Categories are hierarchically organised tags that can be used to filter data in arbitrary ways.

This is an alternative proposal for #908 and should be cleaner for properties that are more list-like. 

CC @henrik242

### Categories

A category is a label organised in a tree-like structure. A label is a simple key word consisting of [A-Za-z0-9_-]. A category defines a label path, a sequence of labels separated by a dot (.). The first label in the path identifies the tree, further labels define the path within the tree. Each category must have at least the root and one node below it. For filtering a maximum of 4 nodes is supported.

You may want to use categories to organise food POIs:

```
food.sit-in.restaurant
food.sit-in.fast-food
food.shop.supermarket
food.shop.convenience_store
```
Or you could express OSM's access hierarchy using categories:
```
access.foot
access.vehicle
access.vehicle.bike
access.vehicle.motor_vehicle
access.vehicle.motor_vehicle.car
...
```

Each place document can be assigned an arbitrary number of categories. Photon doesn't assign any meaning to the categories, with one exception: The `osm` root is reserved and used by Photon to make `osm_key` and `osm_value` searchable.

### Importing categories

You can import categories by using the json import. There already is a `categories` field in the definition which takes an array of strings. The first entry in this field should always be `osm.<osm_key>.<osm_value>`, as Photon uses this category to assign the object's main type. Other than that you can add as many custom categories you like.

The Nominatim export currently only creates the `osm.` category.

### Filtering by category

There are two new API parameters: `include=` and `exclude=`. Use them to filter for categories.

The `include` parameter takes a comma-separated list of categories and will include an object if any of the categories is present. The include parameter may be repeated. The conditions must then be all fulfilled.  So, to restrict the search to drive-thrus, our hypothetical categories from above would filter to:

```
/api?q=...&include=food.sit-in.fast-food&include=access.vehicle
```

The `exclude` parameter also takes a comma-separated list and will exclude and object when **all** categories are present. When repeated, the object will be excluded, when any of the exclude conditions is met.

You can have partial label-paths in your filter condition. `include=food.shop` will output supermarkets and convenience stores. However, you cannot filter by the root label, i.e. `include=food` will (rather surprisingly probably) return nothing.

Category names are currently case-insensitive because of some restrictions with OpenSearch on how synonyms work. That might be a bit surprising because synonyms are not even needed for filtering. However, internally, the categories are also used for category search ("hotel bristol" for finding any hotel named Bristol).

### TODOs

This is just a proof of concept and meant to test out, if it gives the appropriate flexibility for custom filtering. 

Still todo:

- [  ] implement filtering in reverse
- [  ] write tests
- [  ] move `osm_tag` parameter to use our generic filtering instead
- [  ] work around the issue of case-insensitive labels
- [  ] documentation

---

> [!NOTE]
> Adds hierarchical categories to documents, indexes them for filtering and search (include/exclude), replaces legacy classification field, and refactors analyzers/mappings and ingestion accordingly.
> 
> - **Data model**:
>   - Add categories to `PhotonDoc` with validation (`CATEGORY_*`), setters/getters.
> - **Ingestion**:
>   - `NominatimPlaceDocument`: parse `categories` (List), populate `doc.categories`, derive `tagKey/tagValue` from `osm.*`.
>   - `PlaceRowMapper`/`OsmlineRowMapper`: sanitize OSM class/type; set `categories` (e.g., `osm.<key>.<value>`).
> - **Serialization/Indexing**:
>   - `PhotonDocSerializer`: write `categories`; add `#<category>` tokens to `collector.all`; remove `classification` emission.
>   - `IndexMapping`: drop `classification` field; keep collector fields; minor source/mapping tweaks.
>   - `IndexSettingBuilder`: rework analyzers/filters (synonym graphs, multiplexer, classification/extra synonyms); new search/classification analyzers.
> - **Search/API**:
>   - Add category filters to requests (`include`, `exclude`) with validation; plumb through handlers.
>   - `SearchQueryBuilder`: consolidate filters; support include/exclude via new `CategoryFilter`; refactor bbox/layer/osm tag filters.
>   - `OpenSearch*SearchHandler`: pass include/exclude to query builder.
> - **Utils**:
>   - Simplify `buildClassificationString` to `#osm.<key>.<value>` (with invalid-char check).
> - **Tests**:
>   - Update JSON reader and classification tests to use categories and new analyzers.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 400c7208c8eb8bf75f8d2e4fb612253403c656af. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/main/java/de/komoot/photon/opensearch/CategoryFilter.java
+++ b/src/main/java/de/komoot/photon/opensearch/CategoryFilter.java
@@ -1,0 +1,40 @@
+package de.komoot.photon.opensearch;
+
+import org.opensearch.client.opensearch._types.FieldValue;
+import org.opensearch.client.opensearch._types.query_dsl.Query;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class CategoryFilter {
+    private final List<String> categories;
+
+    public CategoryFilter(String categories) {
+        this.categories = Arrays.stream(categories.split(","))
+                .map(s -> "#" + s)
+                .collect(Collectors.toList());
+    }
+
+    public Query buildIncludeQuery() {
+        return Query.of(fn -> fn.terms(t -> t
+                .field("collector.all")
+                .terms(tm -> tm.value(categories.stream()
+                        .map(FieldValue::of)
+                        .collect(Collectors.toList())))
+        ));
+    }
+
+    public Query buildExcludeQuery() {
+        return Query.of(fn -> fn.bool(outer -> outer
+                .mustNot(categories.stream()
+                        .map(s -> Query.of(q -> q.bool(inner -> inner
+                                .must(m -> m.term(t -> t
+                                        .field("collector.all")
+                                        .value(FieldValue.of(s))
+                                ))
+                        )))
+                        .collect(Collectors.toList())))
+        );
+    }
+}
--- a/src/main/java/de/komoot/photon/opensearch/OpenSearchSearchHandler.java
+++ b/src/main/java/de/komoot/photon/opensearch/OpenSearchSearchHandler.java
@@ -46,11 +46,13 @@ public class OpenSearchSearchHandler implements SearchHandler<SimpleSearchReques
     }
 
     private SearchQueryBuilder buildQuery(SimpleSearchRequest request, boolean lenient) {
-        return new SearchQueryBuilder(request.getQuery(), lenient).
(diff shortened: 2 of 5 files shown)


lonvia:
Superseded by #953.

PR #954, opened 2025-10-16
support json output directly from Nominatim dump, without database

gcar-bendingspoons:
When dumping to JSON, support loading the data from a Nominatim dump instead of the database.

The goal of this change is to preprocess the JSON data dumps available from Graphhopper before importing in Komoot search service.

- [ ] Update README with new `-json <output filename> -input-file <nominatim dump filename>`
- [ ] Separate PR for adding the address type to the output

### DIFF
--- a/src/main/java/de/komoot/photon/App.java
+++ b/src/main/java/de/komoot/photon/App.java
@@ -124,16 +124,19 @@ public class App {
         final var dbProps = args.getDatabaseProperties();
 
         try {
-            final var connector = new NominatimImporter(args.getHost(), args.getPort(), args.getDatabase(), args.getUser(), args.getPassword(), dbProps);
-            dbProps.setImportDate(connector.getLastImportDate());
-
             final String filename = args.getJsonDump();
             final JsonDumper jsonDumper = new JsonDumper(filename, dbProps);
-            jsonDumper.writeHeader(connector.loadCountryNames(dbProps.getLanguages()));
-
             final var importThread = new ImportThread(jsonDumper);
+
             try {
-                importFromDatabase(args, importThread, dbProps);
+                if (args.getImportFile() == null) {
+                    final var connector = new NominatimImporter(args.getHost(), args.getPort(), args.getDatabase(), args.getUser(), args.getPassword(), dbProps);
+                    dbProps.setImportDate(connector.getLastImportDate());
+                    jsonDumper.writeHeader(connector.loadCountryNames(dbProps.getLanguages()));
+                    importFromDatabase(args, importThread, dbProps);
+                } else {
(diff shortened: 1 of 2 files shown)


lonvia:
I don't understand what you want to do here. If you need to preprocess the JSON dumps you can do that with any other tool that can process json before importing them to Photon.

gcar-bendingspoons:
Hello @lonvia , thanks for looking into this (I still have to polish the PR though).

Initially, I noticed that the format of the JSON file provided by Graphhopper and the one resulting from running photon with `-json` are not exactly the same. We are revamping the komoot search engine that was built upon the format resulting from running photon with `-json`, and I didn't want to change the parsing too much, so I wanted to use photon directly to obtain the expected format.

Better looking into the files, I found that photon output only adds `osm_key` and `osm_value` (which I just noticed are parsed from the first value of `categories`, can you confirm), and the computed address type (which I added using the field name `type`). It also removes some other fields that we are not using.
The address type should be calculated based on the `rank_address` field.

To perform this preprocessing, from a JSON file to another, I added the possibility to bypass the database, and instead tie together the already implemented functionalities to import from JSON and dump to JSON.

Do you see an issue with adding this functionality?

lonvia:
The JSON dumps on the export server are produced by current Photon. So reading and dumping them again would just be a complicated way of copying the file. But I think I know what is going on here.

Photon had a different form of json dump until this year. That wasn't a very useful dump because it just dumped internal data structures that were prone to being changed and noone every really used that functionality. With possibly the exception of Komoot. They did ask me many years ago for a json dump, which I added to the export server. I never got any feedback after the initial request and the dump creation broke a very long time ago without anybody complaining. When I asked at the beginning of the year, nobody knew about the dumps either, so I assumed that you had dropped the idea of using them after initial experiments. Looks like I was wrong.

So, sorry, the older JSON format you are using is gone and will not come back. Can you please adapt your code to simply use the new format? The changes shouldn't be too big, as you have already noticed. The new format is now part of the official Photon roadmap. That means there will be regular dumps and the format itself is meant to be stable. So much more future proof. There might be a couple more minor changes to the format until we reach Photon 1.0 but after that, it will be considered stable and documented.

Regarding `osm_key` and `osm_value`: you are right that they can be parsed from the `categories` right now. However, I already noticed that it wasn't a good decision to remove the fields and added them back in the latest master version. I also consider exporting the address type. Right now the `rank_address` field has to do.

gcar-bendingspoons:
Ok I see, thanks for the extra information 🙏  I will update our code to parse the new format then. For now I'll copy the code that parses the address type from the rank, so the result is aligned with Photon.

PR #960, opened 2025-10-21
Remove a couple of deprecation warnings after JDK 21 upgrade

henrik242:


lonvia:
Thanks, I've kept meaning to look at those.

PR #962, opened 2025-10-30
Add Micrometer+Prometheus metrics

henrik242:
I hope and believe there is an argument to be made for adding Micrometer+Prometheus in Photon.

I've tried just adding the [Prometheus JMX Exporter](https://prometheus.github.io/jmx_exporter/) but it lacks sorely in things like request and latency metrics.

Other related projects, such as [OpenTripPlanner](https://github.com/search?q=repo%3Aentur%2FOpenTripPlanner%20prometheus&type=code)  also uses the same stack for metrics.

This PR utilizes javalin's own micrometer support and strives to be as nonintrusive as possible.

### DIFF
--- a/src/main/java/de/komoot/photon/CommandLineArgs.java
+++ b/src/main/java/de/komoot/photon/CommandLineArgs.java
@@ -85,7 +85,7 @@ public class CommandLineArgs {
 
     @Parameter(names = "-cors-any", description = "Enable cross-site resource sharing for any origin.")
     private boolean corsAnyOrigin = false;
-    
+
     @Parameter(names = "-cors-origin", description = "Comma-separated list of origins for which to enable cross-site resource sharing.")
     private List<String> corsOrigin = new ArrayList<>();
 
@@ -103,6 +103,9 @@ public class CommandLineArgs {
 
     @Parameter(names = "-import-geometry-column", description = "[import-only] Add the 'geometry' column from Nominatim on import (i.e. add Polygons/Linestrings/Multipolygons etc. for cities, countries etc.). WARNING: This will increase the Elasticsearch Index size! (~575GB for Planet)")
     private boolean importGeometryColumn = false;
+
+    @Parameter(names = {"-metrics-enable"}, description = "Set to 'prometheus' to enable built-in Prometheus metrics endpoint")
+    private String metricsEnable = "";
 
     public String[] getLanguages(boolean useDefaultIfEmpty) {
         if (useDefaultIfEmpty && languages.isEmpty()) {
@@ -228,6 +231,10 @@ public class CommandLineArgs {
 
     public boolean getImportGeometryColumn() {
         return importGeometryColumn;
+    }
+
+    public String getMetricsEnable() {
+        return metricsEnable;
     }
 
     public DatabaseProperties getDatabaseProperties() {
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -3,7 +3,7 @@ photon_data/
 dependency-reduced-pom.xml
 .gradle/
 build
-
+.DS_Store
```
lonvia: I'd prefer if ignores specific to the local setup (MacOS here) go into the `.gitignore` in your home directory. This file should only have the ones specific to the project. (And, yes, the file needs some cleanup but that's for an extra PR.)

src/main/java/de/komoot/photon/metrics/MetricsConfig.java
```diff
@@ -0,0 +1,85 @@
...
+
+    private void registerJvmMetrics(MeterRegistry registry) {
+        new ClassLoaderMetrics().bindTo(registry);
+        new JvmMemoryMetrics().bindTo(registry);
+        new JvmGcMetrics().bindTo(registry);
+        new JvmThreadMetrics().bindTo(registry);
+        new UptimeMetrics().bindTo(registry);
+        new ProcessorMetrics().bindTo(registry);
+        new DiskSpaceMetrics(new File(System.getProperty("user.dir"))).bindTo(registry);
+        new FileDescriptorMetrics().bindTo(registry);
```
lonvia: The latter four look like system-wide metrics. Seems odd to have them exported by Photon.

The current size of the database could be an additional interesting data point, though. It doesn't matter when using a static import but becomes relevant when running updates.

src/main/java/de/komoot/photon/CommandLineArgs.java
```diff
@@ -104,6 +104,12 @@ public class CommandLineArgs {
     @Parameter(names = "-import-geometry-column", description = "[import-only] Add the 'geometry' column from Nominatim on import (i.e. add Polygons/Linestrings/Multipolygons etc. for cities, countries etc.). WARNING: This will increase the Elasticsearch Index size! (~575GB for Planet)")
     private boolean importGeometryColumn = false;
 
+    @Parameter(names = {"-metrics-enable"}, description = "Enable built-in Prometheus metrics endpoint")
+    private boolean metricsEnable = false;
+
+    @Parameter(names = {"-metrics-path"}, description = "Path of Prometheus metrics endpoint")
+    private String metricsPath = "/metrics";
```
lonvia: I think one parameter is enough here, we already have too many of them. :wink: So lets just have a fixed endpoint name.

What we could consider: give the parameter a string value for the metrics type : `-metrics-enable prometheus`. That way, if OTLP takes off and we want to support it later, it will be easy to just offer a different alternative.

lonvia (commented):
Prometheus is more or less state of the art these days for monitoring, so lets go with that.

Ideally, you'd add a simple test to https://github.com/komoot/photon/blob/master/src/test/java/de/komoot/photon/api/ApiIntegrationTest.java to check that the endpoint is actually responding with a HTTP 200, when enabled.

For the record: Adds another 3MB to the jar.

.gitignore
henrik242: Fixed! I usually add it, I've seen others unwittingly check it in.

src/main/java/de/komoot/photon/metrics/MetricsConfig.java
henrik242: Removed.

src/main/java/de/komoot/photon/CommandLineArgs.java
henrik242: Fixed.

henrik242:
I think I've addressed all the things you mentioned.

PR #964, opened 2025-11-05
Add ProcessorMetrics

henrik242:
We rely on CPU metrics from all our other services, so I guess it makes sense here as well.

Ref #962


lonvia:
The documentation of Micrometer is unfortunately not very helpful in explaining what exactly it measures. Is it the CPU usage over all the system or CPU usage of the jvm? I can see a use case for the latter. For the former there must be batteries included with prometheus and I wouldn't want to duplicate that in Photon.

If it is really just for the JVM, then there'd be a case for adding `FileDescriptorMetrics` and `UptimeMetrics` as well.

henrik242:
ProcessorMetrics measure both system-wide (system.cpu.usage) and jvm-specific (process.cpu.usage) usage. It looks like FileDescriptorMetrics and UptimeMetrics are both jvm-specific, so I guess I should add them to the review too.

We need process.cpu.usage, though, so I really hope you'd let it through.

lonvia:
Lets go with the three that have process-specific information, then. Can you change the PR accordingly?

henrik242:
Done!

PR #974, opened 2025-11-13
Add git sha to jar

henrik242:
Having the sha in the jar makes it easier to verify the exact features of Photon

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -105,6 +105,10 @@ tasks.named('jar') {
     manifest.attributes('Multi-Release': 'true')
 }
 
+def gitHashProvider = providers.exec {
+    commandLine 'git', 'rev-parse', '--short', 'HEAD'
+}.standardOutput.asText.map { it.trim() }
+
 shadowJar {
     duplicatesStrategy = DuplicatesStrategy.INCLUDE
     failOnDuplicateEntries = true
@@ -121,6 +125,13 @@ shadowJar {
 
     filesMatching("org/apache/commons/logging/**") {
       duplicatesStrategy = DuplicatesStrategy.EXCLUDE
+    }
+
+    manifest {
+      attributes(
+        'Implementation-Version': version,
+        'Git-Commit': gitHashProvider
+      )
     }
 
     destinationDirectory.set(file('target'))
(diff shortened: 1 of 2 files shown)


mtmail:
Seems useful. I tried to get the same running for Nominatim so that `nominatim --version` returns the git hash.

build.gradle
```diff
@@ -123,6 +127,10 @@ shadowJar {
       duplicatesStrategy = DuplicatesStrategy.EXCLUDE
     }
 
+    manifest {
+      attributes('Implementation-Version': gitHashProvider)
+    }
```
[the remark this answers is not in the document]
henrik242: This is untrue. Gradle unwraps the provider.  (Tested locally)

henrik242:
Cool. It could even be helpful to have it in the /status endpoint, but that's for another PR.

lonvia:
While I like the idea in general, I'd want the released jars to report the official version number instead of the git hash number. Can we either detect that the current git is equal to a tagged version or somehow hand in the version number explicitly?

Alternatively, the 'version' field in the status should report what is the [currently set version](https://github.com/komoot/photon/blob/3375701868226f5007c421aeac95d5bd03316369/build.gradle#L9) and have a separate field 'source_version' (or whatever better name you can come up with) containing the git sha.

henrik242:
Pushed a fix :)

src/main/java/de/komoot/photon/StatusRequestHandler.java
```diff
@@ -21,8 +29,35 @@ public void handle(@NotNull Context context) throws IOException {
...
+    private Attributes readManifestAttributes() {
+        try {
+            var resource = getClass().getResource("/META-INF/MANIFEST.MF");
+            if (resource != null) {
+                try (var stream = resource.openStream()) {
+                    return new Manifest(stream).getMainAttributes();
+                }
+            }
+        } catch (IOException e) {
+            // Ignore, will return unknown values
```
otbutz, line 49:
```suggestion
            throw new UncheckedIOException(e);
```

henrik242: I'd rather just ignore this, e.g. in cases where running in a non-jar environment.

PR #980, opened 2025-12-03
Add query option to always include full addresses with house numbers in result

henrik242:


henrik242:
Just a heads up: We're running this in production now without issues.

src/main/java/de/komoot/photon/opensearch/SearchQueryBuilder.java
```diff
@@ -219,7 +219,24 @@ public void addBoundingBox(Envelope bbox) {
         }
     }
 
+    /**
+     * Include housenumber results that match the query in their parent field.
+     * This allows finding addresses like "Main Street 5" when searching for "Main Street".
+     * Only applies when the query doesn't already contain a house number (digits).
+     */
+    public void includeHousenumbers(String query, boolean includeHousenumbers) {
+        if (includeHousenumbers && !query.matches(".*\\d.*")) {
```
otbutz: I'd suggest extracting the pattern. Otherwise this needs to be compiled for each request.

otbutz:
```java
private static final Pattern HAS_DIGIT = Pattern.compile("^.*\\d.*$");
// ...
    if (includeHousenumbers && !HAS_DIGIT.matcher(query).matches()) {
```

lonvia:
This doesn't quite do what you want. The query that includes the housenumber will now _only_ match housenumber and street. It will ignore everything else in the query. In my test dataset which has a 'Auelestr' in 'Triesen' and 'Vaduz', it will happily return 'Auelstr 16, Vaduz' for the query 'Auelestr Triesen'.

The parameter should rather lift the requirement of the housenumber matching exactly by changing this line https://github.com/komoot/photon/blob/176a0d995e0afb75906545c176dbfc147ec14604/src/main/java/de/komoot/photon/opensearch/SearchQueryBuilder.java#L101

and just replacing the 'must' with a 'should' when the option is set. Possibly amended with a must of "housenumber exists".

I'm also still not convinced about the parameter name. It's not something I could explain to a user. I've not yet come up with a good alternative though.

src/main/java/de/komoot/photon/opensearch/SearchQueryBuilder.java
henrik242: Thanks, resolved now

henrik242:
@lonvia Thanks for the feedback! I've tried to refactor the solution a bit, this should address your findings with `Auelestr Triesen`.  (I've tried finding a better name than  `include_housenumbers` too, without much success.)

src/main/java/de/komoot/photon/opensearch/SearchQueryBuilder.java
```diff
@@ -1,25 +1,31 @@
...
 
 public class SearchQueryBuilder extends BaseQueryBuilder {
     public FunctionScoreQuery.Builder innerQuery = new FunctionScoreQuery.Builder();
     double importance_factor = 40.0;
 
-    public SearchQueryBuilder(String query, boolean lenient) {
+    private static final Pattern HAS_DIGIT = Pattern.compile(".*\\d.*");
+
+    public SearchQueryBuilder(String query, boolean lenient, boolean includeHousenumbers) {
+        boolean effectiveIncludeHousenumbers = includeHousenumbers && !HAS_DIGIT.matcher(query).matches();
```
lonvia: I don't think this extra check is necessary. The query score should take care that the right results float to the top. Also, just looking for a digit will fail as test as soon as you get to the Americas with all their "1st Avenues".

src/main/java/de/komoot/photon/opensearch/SearchQueryBuilder.java
```diff
@@ -97,39 +103,47 @@ public double setupFullQuery(String query, boolean lenient) {
...
+                                                                                .collect(Collectors.toList())
+                                                                ))
+                                                        )
+                                                ))
+                                        .functions(constFact -> constFact.weight(1.0f))
+                                        .build().toQuery();
+
+                                if (includeHousenumbers) {
+                                    b.should(hnrQuery);
+                                    b.must(m -> m.exists(e -> e.field(Constants.HOUSENUMBER)));
```
lonvia, line 130: Looking at the planet results, I'm thinking that we should restrict that to "pure" addresses by requiring that a 'name' doesn't exist. Entries with name and housenumber tend to have a higher importance which makes them float on top in the oddest ways.

henrik242:
Thanks for the thorough feedback! I have switched the parameter to `suggest_addresses`, and also addressed the two changes you suggested. The results still look as expected on my end.

lonvia:
I'm sorry but I strongly disagree with this last change adding the suggest addresses feature to short queries. It will significantly worsen the results because the rest of that query is strongly tuned towards prefix matching.

henrik242:
@lonvia That change is needed here, because I saw differences when searching for e.g. "nils gotlands veg" and "romsdalsveien".  The latter will invoke the short query because of `query.matches("^\\p{IsAlphabetic}+$"`: https://github.com/komoot/photon/blob/176a0d995e0afb75906545c176dbfc147ec14604/src/main/java/de/komoot/photon/opensearch/SearchQueryBuilder.java#L19-L20

 That difference wasn't in [the first commit](https://github.com/komoot/photon/pull/980/commits/50bcb50263ff53119c201326e1575a773c7585f4)

I guess an alternative could be to explicitly exclude suggestAddresses from shortQueries:
```
if (!suggestAddresses && (query.length() < 4 || query.matches("^\\p{IsAlphabetic}+$"))) {
```

lonvia:
> I guess an alternative could be to explicitly exclude suggestAddresses from shortQueries:

That would be better.

henrik242:
> That would be better.

Cool, I just pushed that change.

lonvia:
I'll merge that for now but would consider this still experimental. I'd certainly would like to have feedback from users who will use that, what they use it for.

In the context of the OSM planet DB I can only see it making sense, when doing address-only searches. I'm thinking of adding an additional `address ` option for the 'layers' parameter, which returns the subset of `house` feature that are a full address (aka have a housenumber).

PR #982, opened 2025-12-09
Avoid NPE for nullable score() and source()

henrik242:
`score()` and `source()` are annotated as `@Nullable`

### DIFF
--- a/src/main/java/de/komoot/photon/opensearch/OpenSearchSearchHandler.java
+++ b/src/main/java/de/komoot/photon/opensearch/OpenSearchSearchHandler.java
@@ -34,7 +34,14 @@ public class OpenSearchSearchHandler implements SearchHandler<SimpleSearchReques
 
         List<PhotonResult> ret = new ArrayList<>();
         for (var hit : results.hits().hits()) {
-            ret.add(hit.source().setScore(hit.score()));
+            var score = hit.score();
+            var source = hit.source();
+            if (source != null) {
+                if (score != null) {
+                    source.setScore(score);
+                }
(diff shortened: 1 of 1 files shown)


src/main/java/de/komoot/photon/opensearch/OpenSearchSearchHandler.java
```diff
@@ -34,7 +34,11 @@ public List<PhotonResult> search(SimpleSearchRequest request) {
 
         List<PhotonResult> ret = new ArrayList<>();
         for (var hit : results.hits().hits()) {
-            ret.add(hit.source().setScore(hit.score()));
+            var score = hit.score();
+            var source = hit.source();
+            if (score != null && source != null) {
+                ret.add(source.setScore(score));
```
lonvia: You need to distinguish another cases here: if source is not null but score is, we'd still want to add the source to the result list.

henrik242: That won't work, because setScore takes a `double` (which isn't nullable)

henrik242:
I didn't realize I had to submit the review. Better now?

src/main/java/de/komoot/photon/opensearch/OpenSearchSearchHandler.java
lonvia: Just don't call setScore() when score is null

henrik242: Right, now I finally understood what you meant :)

PR #987, opened 2025-12-17
Trivial cleanup

henrik242:
I have addressed some nullability issues, made use of some more modern jdk features, and removed the mention of ElasticSearch. The changes should be trivial, but please tell if I should split up the PR.

### DIFF
--- a/src/main/java/de/komoot/photon/ConfigExtraTags.java
+++ b/src/main/java/de/komoot/photon/ConfigExtraTags.java
@@ -20,7 +20,7 @@ public class ConfigExtraTags {
     }
 
     public ConfigExtraTags(List<String> tags) {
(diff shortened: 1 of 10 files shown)


henrik242:
I fixed the first cursor comment, and I believe the second (duplicated) comment isn't valid.

lonvia:
Agreed on the Cursor comment. That doesn't make sense. If code is not designed to work with result totaling disabled. So your fix is okay.

Can we shelf the changes to the README for now? I've left the ElasticSearch there intentionally because people come to the README visible on Github for instruction on installing the latest release version and for that it is still important to explain the differences.

The rest looks good.

henrik242:
Sure, the readme changes are gone :)

lonvia:
Thanks, great stuff. There is probably more room for moving code towards Java 21.

PR #997, opened 2026-01-15
Allow empty queries then when categories are present

henrik242:
Makes it possible to use the `include` query parameter without `q` being present.


lonvia:
Tested against a OSM planet and the query remains fast enough even when running against very frequent categories, so we are good here. Thank you.

PR #1000, opened 2026-01-27
Json dump: allow string placeIds and remove fromarea

lonvia:
Some minor changes to the expected JSON dump format:

* The placeID may now be an arbitrary string consisting of letters, numbers, dash and undescore. This makes it easier to mix custom data with the provided OSM dumps without risking to have a clash in the ID space. Dumps now also produce string placeIDs but reading number placeIDs still works, making this change backwards compatible. OpenSearch has always used a string for its document identifier, so there is no change on that side.
* Remove 'fromarea' as an allowable attribute when working with addresslines. This is currently not relevant for dumps produced by Photon.


mtmail:
🎂 PR 1000

PR #1008, opened 2026-02-02
Load geometries as EWKB

### LINKED ISSUE #1007: Improve geometry parsing
https://github.com/komoot/photon/blob/f5a051005edda450b717f97a6c15c9c35498ef61/src/main/java/de/komoot/photon/nominatim/PostgisDataAdapter.java#L36-L45

DB result (binary) -> PGeometry -> WKT -> Geometry

We could simply request `ST_AsBinary(linegeo)` and pipe that into a JTS WKBReader.

otbutz:
Fixes #1007

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -73,7 +73,6 @@ dependencies {
     implementation 'io.javalin:javalin:6.7.0'
     implementation 'io.javalin:javalin-micrometer:6.7.0'
     implementation 'io.micrometer:micrometer-registry-prometheus:1.16.2'
-    implementation 'net.postgis:postgis-jdbc:2025.1.1'
     implementation 'com.fasterxml.jackson.core:jackson-databind:2.21.0'
 
     testImplementation 'com.h2database:h2:2.4.240'


otbutz:
Turns out this is even simpler as the wireformat of Postgis geometries is hex encoded EWKB.

otbutz:
@lonvia using the JTS parser makes the postgis-jdbc dependency obsolete

src/main/java/de/komoot/photon/nominatim/PostgisDataAdapter.java
```diff
@@ -1,24 +1,27 @@
...
 import java.util.HashMap;
+import java.util.HexFormat;
 import java.util.Map;
 
 /**
  * Utility functions to parse data from and create SQL for PostgreSQL/PostGIS.
  */
 @NullMarked
 public class PostgisDataAdapter implements DBDataAdapter {
+    private static final GeometryFactory GEOMETRY_FACTORY = new GeometryFactory(new PrecisionModel(), 4326);
```
lonvia, line 24: We should work with 7-digit precision per default (i.e. `PrecisionModel(10000000)`). Newer versions of Nominatim do cut the precision in the database but who knowns what happens in transfer.

lonvia:
Thanks, it's always good to loose a dependency.

PR #1022, opened 2026-02-16
Add experimental reverse-only mode

### LINKED ISSUE #904: feature request: reverse geocoding only database
In [Dawarich](https://github.com/Freika/dawarich), we only use reverse geocoding, so it doesn't really make sense to have a full database. If possible, it would be great if we would be able to work with the photon database that only contains reverse geocoding data, which will result in the database having smaller size. I think it will be also useful for some other applications as well.

Please let me know if I can assist with it in any way.

Thank you!

lonvia:
Adds a new import parameter `-reverse-only`, which creates a database without any of the indexes for forward search. When running Photon on such a database then the `/api` and `/structured` endpoints are disabled.

A planet database imported as reverse-only comes to about 35GB.

Closes #904.

### DIFF
--- a/src/main/java/de/komoot/photon/App.java
+++ b/src/main/java/de/komoot/photon/App.java
@@ -402,23 +402,25 @@ public class App {
 
         photonServer.get("/status", new StatusRequestHandler(server));
 
-        photonServer.get("/api", new GenericSearchHandler<>(
-                new SimpleSearchRequestFactory(
-                        dbProperties.getLanguages(),
-                        args.getDefaultLanguage(),
-                        args.getMaxResults(),
-                        dbProperties.getSupportGeometries()),
-                server.createSearchHandler(args.getQueryTimeout()),
-                formatter));
+        if (!dbProperties.getReverseOnly()) {
+           photonServer.get("/api", new GenericSearchHandler<>(
+                    new SimpleSearchRequestFactory(
+                            dbProperties.getLanguages(),
+                            args.getDefaultLanguage(),
+                            args.getMaxResults(),
+                            dbProperties.getSupportGeometries()),
+                    server.createSearchHandler(args.getQueryTimeout()),
+                    formatter));
 
-        photonServer.get("/structured", new GenericSearchHandler<>(
-                new StructuredSearchRequestFactory(
-                        dbProperties.getLanguages(),
-                        args.getDefaultLanguage(),
-                        args.getMaxResults(),
-                        dbProperties.getSupportGeometries()),
-                server.createStructuredSearchHandler(args.getQueryTimeout()),
-                formatter));
+            photonServer.get("/structured", new GenericSearchHandler<>(
+                    new StructuredSearchRequestFactory(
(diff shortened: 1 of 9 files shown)


lonvia:
@Freika I don't have good means to test reverse geocoding. So it would be helpful if you or others from Dawarich could test this out. A compiled Photon with the feature will be available for a couple of days from the [actions run](https://github.com/komoot/photon/actions/runs/22058444220).

I've also put a ready-to-use planet database dump at https://download1.graphhopper.com/public/photon-db-planet-reverse-only.tar.bz2 to play around with.

Or you can create a more recent one from the JSON dump for the planet or country extracts with:

```
zstd --stdout -d photon-dump-*.jsonl.zst | java -jar photon.jar import -import-file - -reverse-only
```

Takes around 4.5 hours for the planet to create on a decently spec'd machine.

p0358:
Thank you very much for this, I've tested it with Dawarich there and it seems to work just fine: https://github.com/rtuszik/photon-docker/issues/289

30-35 GB fits nicely on my drive and I've limited the whole Docker container to just 512 MB of RAM. It starts up without problems and returns API responses within like 60 ms. Very happy with not having to dedicate any more RAM or storage to this given the current market.

In this configuration, running Dawarich to re-geocode all points yields 30-40 MB/s SSD read usage. Almost no CPU is used by photon's processes, only sidekiq on the Dawarich side (job scheduler) eats one core, and a bit of Postgres. So this is perfectly reasonable and the bottleneck isn't even in Photon for batching these jobs there. (EDIT: these stats are while Dawarich was still queuing jobs from Postgres into Sidekiq, so some of that disk usage could be just from that)

Also thanks for putting the compressed database dump out there, it surely saved a lot of pain that it'd take to build it from scratch. Though I want to point out that despite the extension being bz2, it looks like you've accidentally compressed it as gzip.

rouke-broersma:
@lonvia would you be able to re-upload the planet db as bz2 instead of gzip so it can be properly unpacked by photon-docker? Would be much appreciated!

lonvia:
That planet db is old by now. Just build your own from the dump files.

lonvia:
And just to be clear: that test dump will not stay around forever. It was really just meant for testing. Please don't code the URL into any docker builds.

binnichtaktiv:
@lonvia Would it be possible to also provide regularly updated prebuilt reverse-only database dumps, similar to the standard Photon databases?

Building the reverse-only database takes several hours and requires more CPU power and RAM than my server (and probably many others) has available.

I tried to build it on my Laptop and after 5 hours it crashed lol

Having an official downloadable reverse-only database would make this feature much easier to use. I know the feature is still marked as experimental, but it is already an official part of Photon, so having official prebuilt reverse-only database dumps would be very helpful

lonvia:
Can't the dawarich community provide their own service for downloading a reverse-only database dump for all? You might make it even easier for everybody and provide them in the form of a pre-loaded docker container. Updating the dump/container once a month would likely be more than enough for that use-case.

PR #1038, opened 2026-03-18
Add OpenAPI spec to docs

PartTimeDataScientist:
I've prepared an OpenAPI file for Photon for the developer documentation of the _MOBIDROM Routing Services_. Might be helpful for users of other Photon instances as well.

Redocly rendering of included OpenAPI spec is [here](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/PartTimeDataScientist/photon/cb27aabe266666c65994e27fa9abcfe3a81b6ea4/docs/openapi.json)

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -1,0 +1,553 @@
+{
+  "openapi": "3.0.1",
+  "externalDocs": {
+    "url": "https://github.com/komoot/photon"
+  },
+  "info": {
+    "title": "Photon Geocoding API",
+    "summary": "photon is an open source geocoder built for OpenStreetMap data. It is based on elasticsearch/OpenSearch - an efficient, powerful and highly scalable search platform.",
+    "version": "1.0.0",
+    "license": {
+      "name": "Apache-2.0",
+      "url": "https://opensource.org/licenses/Apache-2.0"
+    }
+  },
+  "servers": [
+    {
+      "url": "https://photon.komoot.io/",
+      "description": "Public instance sponsored by Komoot"
+    }
+  ],
+  "paths": {
+    "/api": {
+      "get": {
+        "tags": [
+          "Photon Geocoding"
+        ],
+        "summary": "\"Forward\" geocoding: adress -> coordinates",
+        "operationId": "geocoding",
+        "description": "Get coordinates for a given address or POI",
+        "parameters": [
+          {
+            "name": "q",
+            "description": "The `q` parameter contains the term to search for. It is mandatory unless filtering via include/exclude parameters is in place.<br/><br/>",
+            "in": "query",
+            "required": true,
+            "schema": {
+              "type": "string"
+            },
+            "example": "Düsseldorf"
+          },
+          {
+            "name": "limit",
+            "description": "The `limit` parameter sets the number of results - server *may* return less results.<br/><br/>",
+            "in": "query",
+            "required": false,
+            "schema": {
+              "type": "number"
+            },
+            "example": "10"
+          },
+          {
+            "name": "osm_tag",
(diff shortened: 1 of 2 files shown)


docs/openapi.json
```diff
@@ -0,0 +1,553 @@
...
+        "tags": [
+          "Photon Geocoding"
+        ],
+        "summary": "\"Forward\" geocoding: adress -> coordinates",
+        "operationId": "geocoding",
+        "description": "Get coordinates for a given address or POI",
+        "parameters": [
+          {
+            "name": "q",
+            "description": "The `q` parameter contains the term to search for. It is mandatory unless filtering via include/exclude parameters is in place.<br/><br/>",
```
lonvia, line 33: Are these final line breaks really necessary? Shouldn´t spacing rather be fixed in the output renderer?

docs/openapi.json
```diff
@@ -0,0 +1,553 @@
...
+    "version": "1.0.0",
+    "license": {
+      "name": "Apache-2.0",
+      "url": "https://opensource.org/licenses/Apache-2.0"
+    }
+  },
+  "servers": [
+    {
+      "url": "https://photon.komoot.io/",
+      "description": "Public instance sponsored by Komoot"
```
lonvia, line 18: please add here "(for testing and low use only)"

lonvia (commented):
The spec is not completely up-to-date. Some of the new parameters like `include` are missing.

While I'm very open to adding this, it means that we'd have two places with the same documentation. So ideally this comes with a script which converts openapi.json -> api-v1.md.

docs/openapi.json
PartTimeDataScientist: You're right, this could/should be treated by the output renderer. I had added these to improve the Swagger-UI-Rendering for our Docs - but not every renderer might even understand HTML-Code in there. I'll remove them.

grafik

PartTimeDataScientist:
> The spec is not completely up-to-date. Some of the new parameters like `include` are missing.
> 
> While I'm very open to adding this, it means that we'd have two places with the same documentation. So ideally this comes with a script which converts openapi.json -> api-v1.md.

Oh, must have missed that. Understand your point - I thought the api-v1.md to be "more verbose" than an OpenAPI-spec. I'd like to avoid writing a one off-script for this conversion. 

However, I think it could be possible to extend the OpenAPI spec a bit further (although I don't know if it is reasonable to squeeze **all** content from the current api-v1.md into the OpenAPI spec) and then generate a markdown document from it, using e.g.

https://jsontotable.org/openapi-to-markdown
https://www.npmjs.com/package/@scalar/openapi-to-markdown

What do you think? If that sounds reasonable I'll try to extend the OpenAPI file so that it's a reasonable source for a markdown documentation...

lonvia:
I was definitely thinking about a script to regularly use every time the openapi spec changes, not a one-off solution.

> I'll try to extend the OpenAPI file so that it's a reasonable source for a markdown documentation...

Yes, that sounds like a good plan.

I'm not entirely convinced by the npm package you cite, as it claims to focus on LLM-friendly markdown, where I'm looking for human-friendly markdown. But that's rather an implementation detail. This being JSON, it should be fairly straightforward to generate markdown from a template afterwards. https://github.com/vrerv/openapi-markdown looks interesting, as it allows to use custom templates. We could then add any text that doesn't fit with OpenAPI, e.g remarks in the beginning of the file, to the template. Lets think about all that later.

Two more thoughts came up when I skimmed the OpenAPI spec:

It looks like there is a way to [share parameter descriptions]( https://learn.openapis.org/specification/components.html). That would be helpful. Any duplication that can be avoided makes maintenance easier. If you look into the [query package](https://github.com/komoot/photon/tree/master/src/main/java/de/komoot/photon/query), you can see where the source code shares parameter handling.

The other question would be if there is a particular reason you choose json over yaml. I don't have very strong opinions on that and would leave the final choice to whoever does the work. Personally though I find yaml slightly less tedious to edit.

docs/openapi.json
lonvia: It's curious that the OpenAPI spec seems to be rather silent on the question what format to use when adding rich text to the description. I'd expect that it is a requirement for every documentation to provide more than simple text.

arpadgabor:
Can I make a suggestion here? We've been using Scalar for displaying our OpenAPI docs and it's pretty neat, it also renders markdown much nicer than a traditional Swagger.

Maybe adding a static HTML file in the website similar to the [example here](https://scalar.com/products/api-references/getting-started) where the `url` points to the OpenAPI JSON hosted on GitHub would make it easier, as it could always point to the latest JSON on main.

Edit: Ah I see now scalar was mentioned above (actually my suggestion is to just use a scalar page not convert to markdown)

lonvia:
Closing for being stale.

PR #1048, opened 2026-04-10
Optionally use parallel consumers for bulk import indexing

henrik242:
Use a specified number of parallel consumer threads for OpenSearch bulk indexing, optimally one per shard (=5). Each thread has its own Importer instance with an independent BulkRequest buffer, sharing the thread-safe OpenSearchClient.

Replaces the sentinel-based single-thread shutdown with a volatile flag + poll timeout for clean multi-thread termination.

Benchmark (Belgium 11GB, 4.9M docs): 221s -> 166s (25% faster).

PR #1050, opened 2026-04-10
Reduce object allocations in the import hot path

henrik242:
Replace streams and intermediate objects with direct lookups and loops in code that runs per document during import. Main changes: direct Envelope construction for bbox, HashMap for address type lookup, pre-computed array for AddressType.fromRank(), and LinkedHashMap backing for NameMap.

Benchmarked with Belgium (11GB, 4.9M docs, 4 languages): 212s -> 194s (~8% faster) on top of the parallel consumer change.

Needs to use numConsumers == NUM_SHARDS from #1048 to have any effect.

### DIFF
--- a/src/main/java/de/komoot/photon/nominatim/model/AddressType.java
+++ b/src/main/java/de/komoot/photon/nominatim/model/AddressType.java
@@ -24,6 +24,19 @@ public enum AddressType {
     COUNTRY("country", 4, 4, 2),
     OTHER("other", 0, 0, 1);
 
+    private static final AddressType[] RANK_LOOKUP;
+    static {
+        RANK_LOOKUP = new AddressType[31];
+        Arrays.fill(RANK_LOOKUP, OTHER);
+        for (AddressType a : values()) {
+            for (int r = a.minRank; r <= a.maxRank; r++) {
+                if (r >= 0 && r < RANK_LOOKUP.length) {
+                    RANK_LOOKUP[r] = a;
+                }
+            }
+        }
+    }
+
     private final String name;
     private final int minRank;
     private final int maxRank;
@@ -43,10 +56,7 @@ public enum AddressType {
      * @return The corresponding address type or Other if not covered.
      */
     public static AddressType fromRank(int addressRank) {
-        return Arrays.stream(AddressType.values())
-                .filter(a -> addressRank >= a.minRank && addressRank <= a.maxRank)
-                .findFirst()
-                .orElse(OTHER);
+        return (addressRank >= 0 && addressRank < RANK_LOOKUP.length) ? RANK_LOOKUP[addressRank] : OTHER;
     }
 
     public String getName() {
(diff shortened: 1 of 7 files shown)


src/main/java/de/komoot/photon/nominatim/model/NameMap.java
```diff
@@ -28,18 +41,22 @@ public static NameMap makeForPlace(Map<String, @Nullable String> source, Iterabl
     NameMap setLocaleNames(Map<String, @Nullable String> source, Iterable<String> languages) {
         setName("default", source, "_place_name", "name");
         for (var lang : languages) {
-            setName(lang, source, "_place_name:" + lang, "name:" + lang);
+            String[] keys = LOCALE_KEYS.computeIfAbsent(lang,
+                    l -> new String[]{"_place_name:" + l, "name:" + l});
+            setName(lang, source, keys);
```
lonvia: This one I would prefer to revert. I can't imagine that accessing a concurrent hash with all is locking is much faster than concatenating two strings and allocating an array. And it's definitely a lot less readable.

src/main/java/de/komoot/photon/nominatim/model/NameMap.java
```diff
@@ -4,14 +4,27 @@
 import org.jspecify.annotations.Nullable;
 
 import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
 
 @NullMarked
 public class NameMap extends AbstractMap<String, String> {
-    private final Set<Entry<String, String>> entries = new HashSet<>();
+    private final LinkedHashMap<String, String> entries = new LinkedHashMap<>();
```
lonvia: Why do you use a LinkedHashMap here? Why not a simple HashMap? I assume the reason for this change was to have a faster `containsKey()` implementation? Or am I missing something else?

src/main/java/de/komoot/photon/nominatim/model/NameMap.java
henrik242: Yup, I agree. This one was a bit too over-engineered. Fixed.

src/main/java/de/komoot/photon/nominatim/model/NameMap.java
henrik242: The real win was going for a Map instead of a Set, but a HashMap is definitely sufficient. Fixed.

lonvia (commented):
Looks good now. Do you mind squashing the fixup commits together?

src/main/java/de/komoot/photon/nominatim/model/AddressType.java
```diff
@@ -24,6 +24,19 @@ public enum AddressType {
     COUNTRY("country", 4, 4, 2),
     OTHER("other", 0, 0, 1);
 
+    private static final AddressType[] RANK_LOOKUP;
```
otbutz, line 27: Why not a List?

henrik242: It's a tiny fixed lookup table (31 slots) built once at class load. fromRank() runs on every imported row, so a plain array keeps that path as cheap as possible, and since values() already hands back an array, it just felt natural.

henrik242:
> Looks good now. Do you mind squashing the fixup commits together?

Done :)

lonvia:
Thanks. That should already have an effect for the export of the json dump.

PR #1055, opened 2026-04-15
Submit bulk requests in parallel during import

henrik242:
Up to -j N bulk requests are now sent concurrently, with producer backpressure so the pipeline stays bounded.

Bulk writes automatically retry with exponential backoff when OpenSearch pushes back (HTTP 429 from the parent circuit breaker), so large imports don't abort on transient overload.

The index is switched to manual refresh during import and refreshed once at the end, skipping per-second refresh churn.

Belgium benchmark (Apple Silicon, OpenJDK 21, -Xmx4g, 4.95M docs):

                 time     docs/s
    master       233s     21,430
    -j 1         213s     23,260   (-8.6%)
    -j 5         134s     36,970   (-42.5%)

Supercedes #1048

### DIFF
--- a/src/main/java/de/komoot/photon/App.java
+++ b/src/main/java/de/komoot/photon/App.java
@@ -154,10 +154,11 @@ public class App {
             return;
         }
 
-        final var importThread = new ImportThread(esServer.createImporter(dbProperties));
+        final int maxConcurrentBulks = Math.max(1, cli.getGeneralConfig().getThreads());
+        final var importThread = new ImportThread(esServer.createImporter(dbProperties, maxConcurrentBulks));
 
+        Date importDate = null;
         try {
-            Date importDate;
             if (cli.getImportFileConfig().isEnabled()) {
                 importDate = importFromFile(cli.getImportFileConfig(), importFilterConfig, importThread);
             } else {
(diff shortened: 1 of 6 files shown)


lonvia:
Some more background on why I'm currently staling on this:

PhotonDoc may contain more than one place. This is used heavily in the Nominatim data to expand address interpolations. In cities like Buenos Aires with a by-block addressing system, you get thousands of PhotonDocs this way where each document expands to a 100 OpenSearch documents. Thus the effective queue size between the threads is suddenly not at 10_000 but at 1M. That's causing quite a bit of trashing with the garbage collection. And multiple OS threads make this worse.

I'm thinking to change the code so that the import thread simply creates bulk requests directly and puts those in the queue for the OS threads. It shifts a bit of the work (the json assembly) to the import thread but will mean that we safe quite a few intermediate allocations.

henrik242:
Yep, ran into exactly that with Argentina - Buenos Aires is the killer. The latest changes do what you suggested: the producer thread now builds bulk requests directly and hands finished ones to N submitters through a one-slot queue, so the footprint is bounded regardless of how dense an iterable expands. Each producer keeps its own builder so -j N > 1 stays thread-safe.

Argentina (Apple Silicon, OpenJDK 21, -Xmx8g, 4.21M docs):

      -j 1    1464s    2873 docs/s
      -j 5     892s    4715 docs/s   (-39%)

henrik242:
@lonvia Cursor is _finally_ happy. I've squashed the commits so that the PR should be ready for merge if you want it.

lonvia:
Running performance tests on the planet as we speak...

lonvia:
Works well on the planet. The GC warnings are gone. Interestingly there is only a minor performance improvement when running with `-j1`. So the GC trashing wasn't slowing things down.

Running with `-j5` saves a bit more than 2h. I still see the slight bloat though. Might still be worth experimenting with force-merge. But that'd be a different topic.

Thanks for being so patient with Cursor (and me ;).

Issue #550, opened 2021-03-17
Provide OpenAPI specs


debuglevel:
It would be nice if the API provides an endpoint to an OpenAPI specification. This would facilitate code generation for client libraries.

Unfortunately - if Spark did not change much since I used it - it does not support this out of the box. Therefore it would require manual work. There is also https://github.com/manusant/spark-swagger, but I never tried it back then.

KiwiKilian:
Here is the OpenAPI Spec we use https://gist.github.com/KiwiKilian/1f9d35544a7e9312bb74bd75a889df9c.

I don't know anything about spark, therefore I could only contribute the manually created spec. But at least some Swagger UI would be necessary for proper display in the Docs?

danielgit0:
Hi, in the meantime, here is a small update with the status endpoint
[openapi.json](https://github.com/user-attachments/files/15757428/openapi.json)

lonvia:
~There is https://github.com/javalin/javalin-openapi, is somebody want to try their hands on this.~

lonvia:
Given how the AI armies of bots and vibe coders have laid waste on free Internet services over the last year, I'm no longer in favour of having an endpoint that facilitates their work.

I very much like the approach from #1038 where we add a static OpenAPI spec in this repository from which to also generate the human readable documentation.

PR #1058, opened 2026-04-18
Add `/openapi` endpoint using javalin-openapi-plugin

EPNW-Eric:
As requested in #550, this PR adds an OpenAPI endpoint. It does so by exposing the Photon API capabilities as an OpenAPI 3.x document at `/openapi`, using the `javalin-openapi-plugin` community library.

#### `build.gradle`
- Added `io.javalin.community.openapi:javalin-openapi-plugin:7.1.0`: registers the `/openapi` route and generates the JSON spec at startup by reading compile-time metadata.
- Added `io.javalin.community.openapi:openapi-annotation-processor:7.1.0` as an annotation processor; this scans `@OpenApi`-annotated handlers during `compileJava` and emits the metadata the plugin reads at runtime.

#### `App.java`
- Registers `OpenApiPlugin` in the Javalin config with title, version, and description.
- Swaps the three `GenericSearchHandler<T>` route registrations for the new documented wrappers (see below), so the annotated `handle()` methods are what Javalin sees at each path.

#### New handler wrapper classes 
- Each implements `Handler`, delegates to `GenericSearchHandler<T>`, adds a `@OpenApi` annotation to `handle()`
- `ApiSearchHandler`: documents `/api` (forward geocoding): all query parameters including `q`, `lang`, `limit`, `lat`/`lon` bias, `bbox`, `countrycode`, `osm_tag`, `layer`, `include`/`exclude`, `dedupe`, `geometry`, `suggest_addresses`, `debug`.
- `StructuredSearchHandler`: documents `/structured` (structured forward geocoding): address component fields (`city`, `street`, `postcode`, etc.) plus the shared base parameters.
- `ReverseSearchHandler`: documents `/reverse` (reverse geocoding): required `lat`/`lon`, optional `radius`, `distance_sort`, `query_string_filter`, plus the shared base parameters.

#### `StatusRequestHandler`
- Added `@OpenApi` annotation to `handle()` to document the `/status` endpoint.

### Why wrapper classes instead of annotating `GenericSearchHandler` directly

`GenericSearchHandler<T>` is reused for three different routes with different parameter sets. A single `@OpenApi` annotation on a class can only describe one path; the annotation processor links it to exactly one route. Three thin wrapper classes is the idiomatic solution - each has its own `@OpenApi` and delegates to the underlying generic handler, keeping all search logic in one place.

### How to verify

```
./gradlew shadowJar
java -jar target/photon-1.0.0.jar -data-dir <path>
curl http://localhost:2322/openapi | jq .
```

The response should be a valid OpenAPI 3.x JSON document listing `/api`, `/structured`, `/reverse`, and `/status` with their query parameters and response schemas.

---
DISCL**AI**MER:  All code in commit EPNW-Eric/photon@a0f4a940c90b7b71f93f881fe25869b65338d812 as well as this PRs text were written by AI.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ photon_data/
 dependency-reduced-pom.xml
 .gradle/
 build
+bin
 
 .idea/
 *.iml
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,22 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [1.1.0] - 2026-04-18
+
+* add special reverse-only mode creating a smaller DB with on search capability
+* allow arbitrary json data types in extra fields
+* improve handling of postcodes in deduplication
+* add countrycode filter to search (thanks @tsamaya)
+* add rescoring of the OpenSearch result set (adding rematching against query)
+* add postcodes from Nominatim's location_postcodes table to imported data
+* various small performance optimisations (thanks @otbuz, @henrik242)
+* fix dumping of alternative names like alt_name
+* improve scoring of location bias
+
+## [1.0.1] - 2026-03-09
+
+* tolerate adding new properties in future versions
+
 ## [1.0.0] - 2026-02-09
 
 * remove ElasticSearch backend
--- a/build.gradle
+++ b/build.gradle
@@ -73,6 +73,8 @@ dependencies {
     implementation 'org.locationtech.jts.io:jts-io-common:1.20.0'
     implementation 'io.javalin:javalin:7.1.0'
     implementation 'io.javalin:javalin-micrometer:7.1.0'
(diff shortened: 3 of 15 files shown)


lonvia:
Thanks for that. Unfortunately, I've somewhat changed my mind on how to approach the OpenAPI issue and forgot to update #550 accordingly. Right now I'd prefer to go with the approach from #1038, i.e. having a formal OpenAPI spec separate from the service itself. So I'm going to close this.

Apologies for that. Your work is appreciated nonetheless.

EPNW-Eric:
No worries! Just that I understand it correctly: Will there be a formal OpenAPI spec and the code used by photon to expose the service generated based on it? Or will code and spec be decoupled and they both have to be keept in sync manually? Imo having them decoupled is cumbersome, so in my projects I usually write the OpenAPI spec first and by hand and use it as "ground of truth" to later use something like https://openapi-generator.tech/docs/generators/ to generate servercode from it.

PR #1060, opened 2026-04-21
match interior words exactly in short query

henrik242:
For a doc "Marikoven Pøylo", the query "pøylo" returned nothing - instead finding a 1-edit-away "Pøyso" at another location. That fuzzy hit kept the first query non-empty, so the lenient retry that would have found "pøylo" via the interior-word index never ran.

Add a should-clause that exact-matches the interior-word index for queries of 5 chars or more. Boost 0.5 keeps it below the regular prefix match but above the lenient fuzzy fallback.


lonvia:
Nice catch!

Issue #922, opened 2025-07-15
I can not find McDonalds around my current location that I am zoomed in


erik55:
## What did you search for?
McDonalds (when zoomed in to Paris).

## What result did you get?
I got results in somewhere around Paris and maybe some two or three in Paris itself.

## What result did you expect?
I wanted to see the McDonalds restaurants in Paris. It works with OrganicMaps, it works with CoMaps, it works with Google.

**When the result is missing completely:**
The data exists in Openstreetmap.

It is several objects, hundreds as of OrganicMaps.

## Further details
OrganicMaps or CoMaps seem to work as expected.

I know I could also have searched for **mcdonald’s** or **mc donald’s** or **mcdonald's** or **mc donald's** or **mcdonalds** or **mc donalds** or all other variations. It doesn’t matter which of all those strange variations I try, I never get the correct or all results, and especially not only in Paris (but around). With CoMaps/OrganicMaps/Google it just works as expected.

I tried also to search for Markthale 7, didn’t find anything. It also worked with OrganicMaps/CoMaps. Where is this fuzzy?

lonvia:
Seriously?

Closing as duplicate of #356.

erik55:
Sorry, but it also didn’t find other wrongly spelled POIs. This is disappointing.

Like I said in the last paragraph: I tried also to search for Markthale 7, didn’t find anything. It also worked with OrganicMaps/CoMaps. Where is this fuzzy?

PR #1064, opened 2026-05-04
fix "Oslo S" matching random possessive POIs

henrik242:
Searching for "Oslo S" was surfacing things like Tiffany's, Lio's, and
Yaya’s because the analyzer split each `'s` into a bare `s` token,
which prefix_edge_ngram (minGram=1) then made matchable against the
`S` in the query. Stemming `'s` away cleanly via stemEnglishPossessive fixes it.

### DIFF
--- a/src/main/java/de/komoot/photon/opensearch/IndexSettingBuilder.java
+++ b/src/main/java/de/komoot/photon/opensearch/IndexSettingBuilder.java
@@ -254,6 +254,13 @@ public class IndexSettingBuilder {
                         .catenateAll(true))
         ));
 
+        settings.filter("delimiter_name_ngram", f -> f.definition(d -> d
+                .wordDelimiterGraph(w -> w
+                        .preserveOriginal(false)
+                        .stemEnglishPossessive(true)
+                        .catenateAll(true))
+        ));
+
         settings.filter("delimiter_alphanum", f -> f.definition(d -> d
                 .wordDelimiterGraph(w -> w
                         .preserveOriginal(false)
@@ -288,7 +295,7 @@ public class IndexSettingBuilder {
                 .tokenizer("collection_split")
                 .filter("delimited_term_freq",
                         "delimiter_whitespace",
-                        "delimiter_terms",
+                        "delimiter_name_ngram",
                         "name_edge_ngram")
                 .filter(NORMALIZATION_FILTERS)
                 .filter("unique")
--- a/src/test/java/de/komoot/photon/opensearch/PossessiveTokenizationTest.java
+++ b/src/test/java/de/komoot/photon/opensearch/PossessiveTokenizationTest.java
@@ -1,0 +1,91 @@
+package de.komoot.photon.opensearch;
+
+import de.komoot.photon.ESBaseTester;
+import de.komoot.photon.PhotonDoc;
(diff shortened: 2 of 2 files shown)


lonvia:
Given that I've explicitly disabled the possessive removal, there must have been a case where it causes bad results. Should have written a test for it because now I need to remember what exactly it was.

On a more general note: any change to the tokenization creates a database that is not backwards-compatible with older releases. That means that they can only happen with major releases. I've added a label now to that end to mark such changes. I'm not yet sure what's the best way to handle such PRs. For now I'd rather defer them until a new major release is scheduled.

henrik242:
OK. This fixes a real pain for us, though, so we might need to maintain a fork until there's a solution in place. Not ideal, but it is what it is. Hope you figure out why you disabled possessive removal! :)

henrik242:
Is it an option to make this functionality configurable?

lonvia:
> Is it an option to make this functionality configurable?

Not this particular change but normalization steps in general, yes. Let me open an issue with thoughts on that....

henrik242:
> Not this particular change but normalization steps in general, yes. Let me open an issue with thoughts on that....

Ok. I do have [a commit that makes it configurable](https://github.com/komoot/photon/pull/1064/commits/cf30e7a6debca472ef283c514e425de9dddde902) if it ever comes to that. We really don't want to maintain our own fork, that's why we migrated from Pelias :)

lonvia:
I've digged a bit now. Apostrophes handled the way they are because of #922. Basically "McDonald's", "McDonalds", and "McDonald" need to be handed as synonyms for both "McDonald's" and "McDonalds". Fuzzy search can go a long way as long as your data source is consistent in its spelling. When it isn't (and it never is), then results become really funny if not all forms appear in the index. And it becomes a bit more involved when also taking into account all the other uses an apostrophe can have in other languages.

What bothers me most about this particular solution is that it doesn't address the root cause at all but just hacks around one specific use case.  If Norwegian stations would be called 'Trondheim O' and 'Oslo O', we'd now looking into a workaround for not making "O'Learys pub" appear. 

A much simpler hack for the problem at hand would be to always remove the apostrophe, i.e. in this line

https://github.com/komoot/photon/blob/699d4cfa45adfe6e8f6b1608bc157e0eab5ea988/src/main/java/de/komoot/photon/opensearch/IndexSettingBuilder.java#L216

replace with empty string instead of space. This is not a general solution but works for languages with a very restricted apostrophe use for the possessive case and in standing terms inherited from foreign languages. As far as I can see from the data, it should work well with Norwegian.

lonvia:
So I was wrong. The minimum solution would be to have the 'index_name_prefix' analyser only create tokens with and without the possessive s but not the possesive s as a token. For example. "McDonald's" should give us the tokens "mcdonalds" and "mcdonald". That would in fact nice to have in general: the requirement to have the name in the query shouldn't  be triggered on a posessive s or an article.

Sadly I can't seem to come up with any combination of filtering that creates exactly that outcome. Enabling stemEnglishPossesives only leaves 'mcdonald' but not the variant with the s. For 'McDonalds' that wouldn't even be a problem because the analyser is a prefix ngram analyser. However, as it has minimal length 5, the whole thing doesn't work for the "Lio's" example.

henrik242:
Closed in favour of #1077

PR #1066, opened 2026-05-05
Add audit logging

scubbx:
This pull request adds audit logging capabilities. This means, that every incoming request is logged to stdout.
To enable this feature, the parameter `-enable-audit-log` has to be set.

By default, no query content is stored to ensure privacy. To also log the content of each query, the parameter `-audit-log-no-privacy` can be set.

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,11 @@ sure you have a JDK installed. Then run:
 
 This will build and test photon. The final jar can be found in the `target` directory.
 
+## Audit Logging
+
+By default incoming requests are not logged. To enable logging of all incoming requests via
+the API, use the parameter `-enable-audit-log`.
+Requests used for (reverse)geocoding will not contain the actual query, unless the additional parameter `-audit-log-no-privacy` is set.
 
 ## Contributing
 
--- a/src/main/java/de/komoot/photon/config/ApiServerConfig.java
+++ b/src/main/java/de/komoot/photon/config/ApiServerConfig.java
@@ -39,6 +39,18 @@ public class ApiServerConfig {
             """)
     private boolean enableUpdateApi = false;
 
+    @Parameter(names = "-enable-audit-log", category = GROUP, description = """
+            Enable logging of incoming requests. No query parameters are logged
+            for coding requests.
+            """)
+    private boolean enableAuditLog = false;
+
+    @Parameter(names = "-audit-log-no-privacy", category = GROUP, description = """
+            When enabled, all request parameters are logged, even for coding requests.
+            Be cautious, this can contain personal information.
+            """)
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -135,6 +136,11 @@ sure you have a JDK installed. Then run:
 
 This will build and test photon. The final jar can be found in the `target` directory.
 
+## Audit Logging
+
+By default incoming requests are not logged. To enable logging of all incoming requests via
+the API, use the parameter `-enable-audit-log`.
+Requests used for (reverse)geocoding will not contain the actual query, unless the additional parameter `-audit-log-no-privacy` is set.
```
lonvia, line 143: This needs to go into `docs/usage.md`

src/main/java/de/komoot/photon/config/ApiServerConfig.java
```diff
@@ -39,6 +39,18 @@ public class ApiServerConfig {
             """)
     private boolean enableUpdateApi = false;
 
+    @Parameter(names = "-enable-audit-log", category = GROUP, description = """
```
lonvia, line 42: I'd prefer `-enable-query-log`. There is no auditing involved from Photon's side.

lonvia (commented):
Adding simple query logging would be fine but it really should remain as simple as possible:

* use configured logger and stderr
* no privacy mode, when enabled, always logs the full query

More elaborate logging is better done in whatever proxy is used in front of a production Photon instance.

scubbx:
Is "privacy mode" no option at all, or should it just not be the default?

lonvia:
I'd need to be convinced that there is an important use-case that is not covered by the logging provided by the fronting webserver in production.

scubbx:
Thanks for the feedback!
I agree with all of your remarks, except I think the "privacy" feature is of real use:
While sufficient in many scenarios, logging by a proxy (of similar) is not the same as when it is done by the application. When logging via an external component, the e.g. response time is not exactly the one used by only Photon itself. Also, logging directly by the application itself increases debugging capabilities. I think, so far, you agreed with me.
In certain conditions, it is necessary to strip personal information from the logged request, which is part of a geocoding request by necessity. When stripping privacy related content from the request, this logic would have to be implemented inside the corresponding proxy individually, which is prone to errors when repeatedly done for each single setup. Photon itself knows best, which parts of the query are privacy related.
When not providing the possibility to strip personal information from the requests, one cannot use the internal logging capabilities when sensitive information is processed (which is true for many geocoding scenarios).

scubbx:
I have to admit, I found another solution to this problem and would not need this PR to be worked on any further. If it is ok with you, I can delete it?

lonvia:
Alright, no worries.

Issue #856, opened 2024-12-10
Reverse geocoding : multiple languages


ybert:
I would like to use photon to have city and country of POIs.

If I want to save in my DB the result of Photon for the reverse geocoding of my POIs latitude/longitude, I have to make a  query by language to have city and country name in each language supported by my app.
If I manage 10 languages in my application I have to make 10 Photon reverse geocoding calls by POI in my application.

It  would be great to have multiple languages results if asked in the query.

lonvia:
We could have an option similar to `name_details` in Nominatim, which returns all available names. This would also allow to access the currently inaccessible alternative names.

That said, you should be aware that Photon only saves a very limited number of translations. With the current dumps, you can get a maximum of 4 translations. If you are creating your own database, you are likely better off using Nominatim directly for what you are doing.

francesco-laricchia:
> If I manage 10 languages in my application I have to make 10 Photon reverse geocoding calls by POI in my application.

This would be great. I managed to make a custom import from Nominatim with 20 languages and it worked, but it also happens that, in my app's autocomplete, searching for a city located in non-latin-alphabet country, there's not a translation for it in my language ('it') and it returns the results in an alphabet I cannot read, except for a few biggest cities.

For example: "Seoul" in Korea is translated, but Incheon returns "인천광역시".

I was thinking that if photon api returned something like as it follows, I could managed to show the best translation available to the end user...

Example.

lang: it -> "name:it" does not exist -> use "name:en"

```
"type": "Feature",
      "properties": {
        "name": "인천리",
        "name:en": "Incheon",
        "name:ko": "인천리",
        "name:zh": "仁川里",
        "name:ja": "インチョン",
        "name:ru": "Инчхон",
        "name:de": "Incheon",
        "osm_type": "N",
        "osm_key": "place",
        "osm_value": "city",
        // ... other properties
      }

```




As an alternative, being able to specify a "lang" and an "alt_lang" in the query parameters would solve the problem:

```
"lang": "it"
"alt_lang": "en"
```

That would give me the english translation instead of the default one which could be in any alphabet in the world actually. Fallback to the local name only if also "alt_lang" is not available.

PR #1067, opened 2026-05-11
Add fallback language

tjvanderende:
At campercontact we use photon for our main search. When someone searches a foreign country, without any translations available for the provided `lang` key Photon defaults to the local language. Most of the times this is works very well, but for non-latin alphabet countries this gives some challenges. 
For example: 
image
A similar issue is described here: https://github.com/komoot/photon/issues/856.

The PR we propose is to include a 'fallback_language' as configurable server variable. This will instead of local language default to the fallback language. It's not yet the perfect solution, but it solves the problem for us - so it might be relevant for others as well.


maglicheev-check24:
@tjvanderende Wouldn't it be better to put this fallback language as a query parameter, instead of one defined on startup? Or even support [logic like Nominatim's](https://nominatim.org/release-docs/latest/api/Search/#language-of-results), where one can pass a list of languages in order of preference.

lonvia:
This is explicitly about the server defaults, not about per-query settings. We'll probably will go for preference lists at some point but for now this will do.

@tjvanderende Looking at your code I'm thinking that we don't even need the additional parameter. It should be sufficient to use the `-default-language` as the fallback before going to default. Or could you foresee a situation where one would set `-default-language` and `-fallback-language` differently? It would the code slightly simpler.

tjvanderende:
As I understood the default language is there only when you didn't include a `lang` parameter. 
I added `fallback-language` to have a back-up for when `lang` is set, but photon wasn't be-able to find the translation. This is a different use-case than default-language.
If we use `default-language` I would need to change how it behaves, which I didn't want to.

The situation right now is that we always have a `lang` parameter included, so we don't use the `default-language`. But I could imagine because we have a dutch application, we would set the `default-language` to `nl` for in the cases we didn't include `lang`, and return `nl` by default. But still default to `fallback-language` to `en` for instances where the dutch translation isn't available.

Do you have suggestions on how we can use the default-language?

lonvia:
`--default-language` describes the language to use, when a specific request does not contain any language requests. `--fallback-language` would set the language to be tried when none of the languages specified in the request matches. I can't make out any situations where you would set these two parameters to different languages. So I'd have no issue extending the meaning of the `--default-language` parameter a bit from:

> language returned when the request has no other language specified

to 

> language returned when the request has no other language specified or none of the specified languages matches

It just seems to be a logical extension of the meaning of default language.

For this PR it would simply mean getting rid of the additional parameter definition and using `args.getDefaultLanguage` where `args.getFallbackLanguage` is used right now.

tjvanderende:
it would simplify it by a lot, thanks! Will do an update this week

tjvanderende:
@lonvia can you check the implementation again? I removed `fallbackLanguage` and instead use the default language.

src/main/java/de/komoot/photon/query/RequestFactoryBase.java
```diff
@@ -73,13 +74,10 @@ protected void completeBaseRequest(RequestBase request, Context context) {
...
-        final String langParam = context.queryParamAsClass("lang", String.class)
-                .check(l -> l == null || "default".equals(l) || supportedLanguages.contains(l),
-                        "Language is not supported. Supported are: default, "
-                                + String.join(", ", supportedLanguages))
-                .getOrNull();
+        final String langParam = context.queryParam("lang");
 
-        if (langParam != null) {
+        if (langParam != null
+                && ("default".equals(langParam) || supportedLanguages.contains(langParam))) {
```
lonvia: This change seems unrelated and change the behaviour of Photon from a strict parameter check to ignoring unknown languages.

src/main/java/de/komoot/photon/searcher/GeoJsonFormatter.java
```diff
@@ -130,6 +131,12 @@ public String convert(List<PhotonResult> results, String language,
         return writer.toString();
     }
 
+    public String convert(List<PhotonResult> results, String language,
+                          boolean withGeometry, boolean withDebugInfo,
+                          @Nullable String queryDebugInfo) throws IOException {
+        return convert(results, language, null, withGeometry, withDebugInfo, queryDebugInfo);
+    }
```
lonvia: This looks to be only used in tests now. Can you please remove this wrapper and simply adopt the tests where convert() is called to take another parameter?

lonvia (commented):
Can you please rebase your branch on master and then force-push the result?

Bonus points for also rearranging the commits to get rid of the fixup commits (although this might be a bit tedious in this case, I admit).

tjvanderende:
Thnx for the feedback @lonvia . Both the commits & suggestions are fixed now. Can you check them again?

src/main/java/de/komoot/photon/query/RequestFactoryBase.java
```diff
@@ -73,11 +74,7 @@ protected void completeBaseRequest(RequestBase request, Context context) {
     }
 
     private String parseLanguage(Context context) {
-        final String langParam = context.queryParamAsClass("lang", String.class)
-                .check(l -> l == null || "default".equals(l) || supportedLanguages.contains(l),
-                        "Language is not supported. Supported are: default, "
-                                + String.join(", ", supportedLanguages))
-                .getOrNull();
+        final String langParam = context.queryParam("lang");
```
lonvia: Your revert didn't work quite right here. You'd also want to reinstate the check().

tjvanderende:
Sorry was to fast with my fix - I reverted it back to the initial state.

lonvia:
bugbot run

lonvia:
Looks good now, thank you.

PR #1077, opened 2026-05-18
fix "Oslo S" matching possessive POIs

henrik242:
Strip "'s" only in the index_name_ngram path via a new delimiter_name_ngram filter. delimiter_terms (term collector) stays unchanged so "Lio's" still hits a perfect match.

### DIFF
--- a/src/main/java/de/komoot/photon/opensearch/IndexSettingBuilder.java
+++ b/src/main/java/de/komoot/photon/opensearch/IndexSettingBuilder.java
@@ -275,6 +275,12 @@ public class IndexSettingBuilder {
                         .catenateAll(true))
         ));
 
+        settings.charFilter("expand_possessive", f -> f.definition(d -> d
+                .patternReplace(p -> p
+                        .pattern("(\\p{L}+)'s\\b")
+                        .replacement("$1 $1s"))
+        ));
+
         settings.filter("delimiter_alphanum", f -> f.definition(d -> d
                 .wordDelimiterGraph(w -> w
                         .preserveOriginal(false)
@@ -306,7 +312,7 @@ public class IndexSettingBuilder {
         ));
 
         settings.analyzer("index_name_ngram", f -> f.custom(d -> d
-                .charFilter("normalize_apostrophes")
+                .charFilter("normalize_apostrophes", "expand_possessive")
                 .tokenizer("collection_split")
                 .filter("delimited_term_freq",
                         "delimiter_whitespace",
--- a/src/test/java/de/komoot/photon/TestServer.java
+++ b/src/test/java/de/komoot/photon/TestServer.java
@@ -6,6 +6,7 @@ import de.komoot.photon.opensearch.PhotonIndex;
 import de.komoot.photon.searcher.PhotonResult;
 import org.codelibs.opensearch.runner.OpenSearchRunner;
 import org.opensearch.client.opensearch.core.search.Hit;
+import org.opensearch.client.opensearch.indices.analyze.AnalyzeToken;
 
 import java.io.IOException;
 import java.util.List;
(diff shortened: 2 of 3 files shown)


lonvia:
This version only creates the names with the possessive completely removed. You can see this if you run the analyser directly. Create a file `analyzer.txt` like this:

```
{
    "analyzer": "index_name_ngram",
    "text": "lio's"
}
```

Then make sure photon is running and run:

```
curl -XPOST "http://localhost:9201/photon/_analyze" -H 'Content-Type: application/json' -d@analyze.txt | jq
```

Before the change, the output would have `lio`, `s` and `lios` as tokens. After the change it has only `lio`. What I'm actually after is an output of `lio` and `lios`. I can't seem to come up with a combination of token filters which does exactly that.

henrik242:
@lonvia I tried to address it in the latest commit, is this a viable path?

EDIT: Did another update, works fine locally:
```
$ curl -s -XPOST "http://localhost:9201/photon/_analyze" -H 'Content-Type: application/json' -d@analyze.txt | jq
{
  "tokens": [
    {
      "token": "lio",
      "start_offset": 0,
      "end_offset": 3,
      "type": "word",
      "position": 0
    },
    {
      "token": "lios",
      "start_offset": 0,
      "end_offset": 5,
      "type": "word",
      "position": 1
    }
  ]
}
```

src/main/java/de/komoot/photon/opensearch/IndexSettingBuilder.java
```diff
@@ -275,6 +275,23 @@ private void addDefaultSettings() {
...
+                .wordDelimiterGraph(w -> w
+                        .preserveOriginal(false)
+                        .stemEnglishPossessive(true)
+                        .catenateAll(true))
+        ));
+
+        // Pre-tokenize possessive forms so "Lio's" indexes as both "lio" (via
+        // stemEnglishPossessive in delimiter_name_ngram) and "lios" (the
+        // catenated form). Doing this as a char filter avoids the offset
+        // violations a multiplexer would introduce around word_delimiter_graph.
```
lonvia: A general comment here: These comments are valuable for understanding the code changes but in my experience tend to get stale because they aren't maintained, at least not by human coders. How about moving them into the commit message as an extended comment. That way it is easy to dig them up again with `git blame` if required but not have them around forever in the code.

src/main/java/de/komoot/photon/opensearch/IndexSettingBuilder.java
```diff
@@ -275,6 +275,23 @@ private void addDefaultSettings() {
...
+        settings.charFilter("expand_possessive", f -> f.definition(d -> d
+                .patternReplace(p -> p
+                        .pattern("(\\p{L}+)'s\\b")
+                        .replacement("$1's $1s"))
```
lonvia: With that little trick you don't even need the `stemEnglishPossesive` special-casing anymore if you use a replacement of `$1 $1s`.

lonvia (commented):
Nice! Works well even with the importance hack and collection split filters

I'm naturally now thinking that we'd want the same to work with "O'Learies" and "L'eglise". But maybe I'm getting ahead of myself. Let's check out first what other languages doing with the apostrophe.

src/main/java/de/komoot/photon/opensearch/IndexSettingBuilder.java
henrik242: sure! I'll move it.

src/main/java/de/komoot/photon/opensearch/IndexSettingBuilder.java
henrik242: will fix

henrik242: Done. I could even ditch the `delimiter_name_ngram` now 🎉

PR #1095, opened 2026-07-07
Restore HTTP request-latency histogram buckets in Prometheus metrics

henrik242:
The metric filter still matched the old jetty.server.requests timer, but Javalin renamed it to http.server.requests, so percentile histograms were never enabled and the http_server_requests_seconds_bucket series vanished. Match the current name and add regression tests (filter-level plus an end-to-end scrape).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,16 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+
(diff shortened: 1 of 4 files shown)


lonvia:
Claude's chattiness (I presume this is Claude) is becoming a problem. It takes me twice as long to read the comments as it takes to read the code they comment. Could you instruct it to keep comments to a minimum and have a single comment not exceed 2 lines?

henrik242:
Haha, it's indeed very chatty. I've already tuned it down quite a bit, but it seems I need to tighten that thumbscrew some more.

PR #1097, opened 2026-07-07
Tweak query rematching to put larger weight on name match

lonvia:
This tweaks the rematching algorithm to put more weight on full name matches. In detail, it changes the following:

* Name matching is done before address matching and gets a larger weight than a matching address.
* Generally reduce match score for partial matches.
* When a housenumber is present, then the primary matching is against the street name rather than the name of the object.
* Country code and country name matches are only considered if they match completely and appear at the beginning or end of the query.

@henrik242 Could you give this a try to see if there are any regressions?


henrik242 (approved):
@lonvia No regressions that I can detect. Looks solid 👍🏻

lonvia:
@cursor review

PR #1109, opened 2026-08-14
update dependencies

henrik242:


settings.gradle
```diff
@@ -1,2 +1,4 @@
+plugins {
+    id 'io.github.ben-manes.versions.settings' version '0.61.0'
```
lonvia, line 2: What's the reason for moving this into settings? Should we move all plugins into settings?

henrik242: @lonvia No, it's just the new recommended way of applying this specific plugin: https://github.com/ben-manes/gradle-versions-plugin#applying-the-plugin

PR #1115, opened 2026-08-17
homepage: remove unused JS dependencies and fonts

mtmail:
Removes 14 files and a bunch of small changes
* removes javascript dependencies for Angular and Underscore, they were not used
* removes old font type of unsupported browsers `#iefix`. Then turns out the font (copyright Google 2006) isn't even listed on fonts.google.com anymore and the browser default font looks similar enough so removed all font files
* removes solr.svg because is not linked anywhere
* removes the email address. It was already commented out
* best practice to treat even OSM data as text instead of HTML. Leaflet's `.bindPopup` inserts HTML by default
* best practice to use SRI hashes when loading libraries from CDNs


lonvia:
Thanks.

PR #1122, opened 2026-09-04
Improve synonym loading

### LINKED ISSUE #1070: Update synonyms without rewriting index settings
When configuring synonyms, then Photon rewrites the index mappings from scratch. This ties the index layout to the Photon version instead of the current database import. This in turn restricts changes to the layout and analysers to major Photon versions. It also prevents customization of the analysers, for example, to adapt to the language of the data.

We need to be able to set and change synonyms without affecting the index mapping. Two possible ideas for a solution:

* Move synonyms and classification terms out of OpenSearch and do the preprocessing ourselves. Probably means that we need to create multiple queries for different interpretations of the query, a bit like Nominatim does.
* Always have a synonym and classification filter and only change the synonym terms on startup. This means reading the database layout description from the database, changing the relevant fields and writing the configuration back.

lonvia:
This changes how synonyms are loaded when serving: instead of rewriting the entire analyser configuration only the filters with the synonyms are updated.

Leaving the remaining analyser configuration untouched means that we can introduce breaking changes into the configuration. Such changes will now only be effective for new installations while existing ones can keep the analysers that are compatible with their database.

The change is not completely backwards compatible because the new installation method expects the synonym filters to be always present in the search analysers while the old method would only install them, when needed. Enabling synonyms on databases imported before this change will only work, if synonyms had already been enabled. The good news is that you can easily do that by starting an older version of Photon with synonyms. There is an error message which explains the steps.

Closes #1070.

### DIFF
--- a/src/main/java/de/komoot/photon/DatabaseProperties.java
+++ b/src/main/java/de/komoot/photon/DatabaseProperties.java
@@ -24,6 +24,10 @@ public class DatabaseProperties {
     private boolean supportStructuredQueries = true;
     private boolean supportGeometries = false;
     private boolean synonymsInstalled = false;
+    // Transitory attribute that checks that the version supports the
+    // newer installation method of synonyms. Can be removed when
+    // database version becomes greater than 1.0.0-4.
+    private boolean synonymFiltersAvailable = false;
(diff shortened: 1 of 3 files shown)


lonvia:
bugbot run

PR #1123, opened 2026-09-06
metrics: document the OpenSearch gauge semantics in HELP text

unseenmagik:
The OpenSearch gauges in `OpenSearchMetrics` are registered without a description, so `/metrics` exposes no `# HELP` line for any of them.

This matters most for `opensearch_cluster_health_status`, which encodes `2 = green, 1 = yellow, 0 = red` — the inverse of the `elasticsearch_exporter` convention, where a higher value means worse health. A healthy cluster reports a constant `2` indefinitely, so a scrape is indistinguishable from a gauge stuck at a bad value, and nothing in the exposition states the scale. This has already produced a false "permanently RED cluster" report downstream (rtuszik/photon-docker#374).

Changes:

- `.description(...)` on all ten OpenSearch gauges
- a regression test asserting the health status encoding appears in the HELP text

The numeric mapping itself is unchanged — flipping it would break dashboards that already read it correctly.

Verified with `./gradlew test --tests de.komoot.photon.api.ApiMetricsTest` (4/4 pass). Reverting just the health gauge description fails the new test and leaves the other three passing, confirming it pins the HELP text rather than passing incidentally.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


lonvia (commented):
Please remove all Claude comments in the code. They add no value.

lonvia:
@henrik242 Do you mind checking if that looks okay for you?

And should we switch the health status meanings to align with [elasticsearch-exporter](https://github.com/prometheus-community/elasticsearch_exporter) to avoid further confusion?

lonvia:
I'm closing this as per our AI contribution policy. Claude-generated PRs need to have a human behind them that takes responsibility for the code.

PR #1131, opened 2026-09-14
fold diacritics and credit synonym terms when reranking

henrik242:
Two fixes to `QueryReranker`, which compares the query and the result name as plain strings.

1. Fold diacritics before comparing, using Lucene's `asciifolding` (the same filter the index chain applies), and treat `.` as a word break. An ASCII query could not match a name carrying diacritics: `bo sykehjem` now matches `Bø sjukeheim`.

2. Canonicalise every term of a `search_synonyms` class to one representative, on both sides, so a synonym hit earns the points a literal match earns. No-op unless `-synonym-file` is given.

Measured against a full Norwegian index (150k names), top-hit scores:

| query | before | after |
|---|---|---|
| `karl johans gt` | 1.34 | 2.00 |
| `askvoll kirke` | 1.26 | 2.00 |
| `bo sykehjem` | 1.00 | 2.00 |


src/main/java/de/komoot/photon/opensearch/IndexSettingBuilder.java
```diff
@@ -76,6 +76,11 @@ public IndexSettingBuilder setSynonymFile(@Nullable String synonymFile) throws I
         return this;
     }
 
+    @Nullable
+    public List<String> getSearchSynonyms() {
+        return synonymConfig == null ? null : synonymConfig.getSearchSynonyms();
+    }
+
     private void updateSynonymFilters() {
         final var synonyms = synonymConfig == null ? null : synonymConfig.getSearchSynonyms();
```
lonvia, line 85: Use the new function?

src/main/java/de/komoot/photon/searcher/QueryReranker.java
```diff
@@ -2,32 +2,65 @@
...
+    private final Map<String, String> synonyms;
     private final String query;
     private final String language;
     @Nullable private final String fallbackLanguage;
     private final boolean isMultiTermQuery;
     private final boolean isFullQuery;
 
     public QueryReranker(String query, String language, @Nullable String fallbackLanguage) {
+        this(query, language, fallbackLanguage, Map.of());
+    }
```
lonvia, line 29: If this variant of the constructor is preserved for tests only, then I prefer to change the tests and get rid of it.

src/main/java/de/komoot/photon/opensearch/OpenSearchSearchHandler.java
```diff
@@ -20,10 +21,16 @@ public class OpenSearchSearchHandler implements SearchHandler<SimpleSearchReques
     private static final double NEG_DECAY_FACTOR = Math.log(0.5);
     private final OpenSearchClient client;
     private final String queryTimeout;
+    private final Map<String, String> synonyms;
 
     public OpenSearchSearchHandler(OpenSearchClient client, int queryTimeout) {
+        this(client, queryTimeout, Map.of());
+    }
```
lonvia, line 28: If this variant of the constructor is preserved for tests only, then I prefer to change the tests and get rid of it.

lonvia (commented):
Diacritics folding is definitely not going to work this way because it completely reverts what the reranker was supposed to do: it perfectly matches a query input of "bø" against a name "bo". The ranker must only match between unnormalized query and normalized result name, not the other way around. Ideally there'd still be a slight penalty when such a combination matches versus an exact match.

What complicates things: we now have configurable normalizers in place. If you run a `german_normalization` normalizer, you'd also might want the reranker to match 'oe' -> 'ö' (but not 'ö" -> 'oe').

---

Taking synonyms into account is indeed necessary. The implementation looks good to me minus the small nitpicks mentioned.

lonvia:
I've opened a PR which includes the requested changes in #1134. Closing here.

PR #1132, opened 2026-09-14
keep synonyms out of the short-query fallback

henrik242:
A one-token query goes through `setupShortQuery`. Its strict clauses use `search_prefix`, which
has no synonyms, so they often find nothing and the lenient retry runs. That one is analyzed with
`search`, where the synonym filter sits, so the query expands through its whole class and matches
every member of it.

With `gt,gt.,gate,gata` installed, a bare `gt` returned every street named `... gate` in the
country, ordered by importance. It now returns `Nykirken i C Sundts gt` again. `karl johans gt`
is unchanged.

The lenient clause now uses `search_nosyn`, a copy of `search` without the synonym filter.

`DATABASE_VERSION` is bumped because `updateIndexSettings` pushes only filter definitions, not
analyzers, so `search_nosyn` exists only after an import. Without the bump a new jar on an old
index would reference a missing analyzer and fail every short query. **Needs a reimport.**


src/main/java/de/komoot/photon/Server.java
```diff
@@ -38,7 +38,7 @@ public class Server {
      * changes in an incompatible way. If it is already at the next released
      * version, increase the dev version.
      */
-    public static final String DATABASE_VERSION = "1.0.0-4";
+    public static final String DATABASE_VERSION = "1.0.0-5";
```
lonvia, line 41: I don't think this is necessary as only the analyser on the search side is changed. They both work on the same database.

lonvia (commented):
It is an interesting question if the better approach here would be to remove the synonyms for the lenient part or add them in the rest of the short search queries. Did you experiment with that?

lonvia:
Turns out that the direction was right but the solution was incomplete: both synonyms and classification terms should be disabled. That in turn makes the task a lot easier because we can get rid of the multiplexer. I've added a revised version to #1134. Closing here.

Issue #97, opened 2014-08-14
Handle abbreviations in typed names


RicoElectrico:
By the convention, names in OSM database are entered in their full name. So we have "Avenue" instead of "Ave", "Generała" instead of "Gen." and so on. This makes results not necessarily accurate when the user enters the abbreviated form.
A good starting point would be the list from this program which is used in Osmapa.pl Mapnik style to shorten (e.g. street) names that don't fit on a low zoom level. 
https://github.com/balrog-kun/shrtnms/blob/master/shorten.c
It's under 2-clause BSD license, so should be compatible.

lonvia:
The plan here would be to implement the abbreviation definition system that Nominatim uses already, see [Token analysis documentation](https://nominatim.org/release-docs/latest/customize/Tokenizers/#token-analysis). This seems to be doing a decent job and means that we could simply reuse the configuration already established with Nominatim.

Variants and mutations should both be implemented because they both have their uses.

henrik242:
Worth knowing before implementing variants, since they would introduce multi-token rules: `search_synonyms` silently drops query tokens when a rule has a multi-word left-hand side.

On master (`82538833`), default `-normalization-filters lowercase,asciifolding,german_normalization`, with `{"search_synonyms": ["nordre aasen,nordreaasen"]}`:

    _analyze, "search" analyzer, input "nordre aasen"
      without the rule:  [nordre, aasen]
      with the rule:     [aasen]

`nordre` is gone and the synonym is never produced. The other direction is fine (`nordreaasen` -> `[nordre, nordreaasen, aasen]`), and single-token rules are unaffected.

It is the multiplexer rather than the graph filter. Same filters, same input:

    synonym_graph alone                       -> nordreaasen@0, nordre@0, aasen@1
    synonym_graph + word_delimiter_graph      -> nordreaasen@0, nordre@0, aasen@1
    inside multiplexer_search (two branches)  -> aasen

`MultiplexTokenFilter`'s shared selector desyncs once a branch consumes more tokens than it emits, which is what a multi-token LHS does. `classification_terms` already rejects multi-word entries in `IndexSettingBuilder`; `search_synonyms` has no equivalent check, so it fails silently.

lonvia:
There is a warning in the [synonyms doc](https://github.com/komoot/photon/blob/master/docs/synonyms.md) about this to be fair. I've experimented with this when switching to OpenSearch and concluded that four major versions later synonym handling in OS is still severely lacking.

[...]

PR #1134, opened 2026-09-17
Make reranking aware of synonyms

lonvia:
This builds on #1131 but removes the normalization and reorganises the code a bit.

Spaces in synonyms are now already rejected on startup (see https://github.com/komoot/photon/issues/97#issuecomment-5678482194) as they will also cause havoc with the search query.  Also the code is to get the synonyms into the reranker is juggled around a bit.

Otherwise uses the same algorithm as #1131 to switch all synonym words to a canonical form.


lonvia:
Also added a variant of #1132 here, as it turns out that they were more related than initially thought: synonym/classification expansion is now disabled for one-word queries.

@cursor review

lonvia:
bugbot run

henrik242 (approved):
lgtm

